PR #7, opened 2016-11-25
New Avro Schemas Heart Rate

herkulano:
- descriptive_statistic.avsc is the set of available statical value
 - effective_time_frame.avsc describes a time window
 - heart_rate_dataset.avsc states a hear rate dataset
 - heart_rate_.avsc states a single hear rate sample
 - heart_rate_item.avsc is this the dataset item
 - heart_rate_value.avsc represents the heart rate double value
 - unit.avsc is the set of all available units

### DIFF
--- a/RADAR-Questionnaire/config.md
+++ b/RADAR-Questionnaire/config.md
@@ -1,20 +1,18 @@
 ## Proposal for configuration schema
 
 ### Metadata
-```json
+```js
 {
   "version": "0.0.1",
-  "lastModified": "2016-11-18T23:01:01+02:00",
+  "lastModified": "2016-11-18T23:01:01+02:00", // format: ISO8601 or Unix Epoch Timestamp
   "id": "PHQ8",
   "title": "Weekly Checkup",
   "questions": []
 }
 ```
-**Notes:**  
-`"last_modified"` format: ISO8601 or Unix Epoch Timestamp
 
 ### Question type: "Radio"
-```json
+```js
 {
   "id": "PHQ8-1",
   "lead": "Over the past two weeks, how often have you been bothered by any of the following problems?",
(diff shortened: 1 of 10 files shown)


herkulano:
Please review the heart-rate data schema. I added an example of how the data looks like following the OMH standards as @fnobilia has already put it in the AVRO schema files.

https://github.com/RADAR-CNS/RADAR-Schemas/blob/fe41262a38a2c6610e87a3611ed41e5b5ff1f250/REST-API/Heart%20Rate/README.md

fnobilia:
To query REST API you need to provide userID and sourceID. Since this information must been known in advance I thought that this block
`"header": {
    "id": "123e4567-e89b-12d3-a456-426655440000",
    "creation_date_time": "2013-02-05T07:25:00Z",
    "acquisition_provenance": {
      "source_name": "RunKeeper",
      "source_creation_date_time": "2013-02-05T07:25:00Z",
      "modality": "sensed"
    },
    "user_id": "user1432"
  },`
was redundant. While the `user_id` is directly specified, the `acquisition_provenance` informations can be extracted by another request. Clearly we pay the cost of an additional request but we reduce the amount of data sent out upon each interaction. The idea of removing redundant data was discussed with @afolarin and Robert Sugar during one of our first meeting. 

If you check the the dataset schema, you will not find an equivalent in OMH repository. It has been defined mixing FitBeat and OMH schemas since OMH does not have a schema for data set.

PS: for this reason I replied to another comment saying 
> so far we have a custom solution based on OMH. Anyway it is not the final one.

herkulano:
@fnobilia agree. the lack of a dataset in OMH is restrictive and as you say completely redundant for our cases.

herkulano:
@fnobilia does this look closer to what you defined in AVRO?

```json
{
  "header": {
    "descriptive_statistic": "average",
    "unit": "beats/min"
  },
  "dataset": [{
    "heart_rate": {
      "value": 50
    },
    "effective_time_frame": {
      "date_time": "2013-02-05T07:25:00Z"
    }
  }, {
    "heart_rate": {
      "value": 50
    },
    "effective_time_frame": {
      "date_time": "2013-02-05T07:25:00Z"
    }
  }]
}
```

fnobilia:
@herkulano The current AVRO is
`{"dataset":[
	{
	"org.radarcns.avro.HeartRateItem":
		{
		"heart_rate":0.6117274154968215,
		"effective_time_frame":	
			{
			"start_date_time":"2016-10-27T20:02:20Z",
			"end_date_time":"2016-10-27T20:02:30Z"
			}
		}
	},
	{
	"org.radarcns.avro.HeartRateItem":
		{
		"heart_rate":108.39180563508566,
		"effective_time_frame":
			{
			"start_date_time":"2016-10-27T20:03:00Z",
			"end_date_time":"2016-10-27T20:03:10Z"
			}
		}
	}],
"effective_time_frame":
	{
	"start_date_time":"2016-10-27T20:02:20Z",
	"end_date_time":"2016-10-27T20:03:10Z"
	},
"descriptive_statistic":"average",
"unit":"beats_per_min"
}`
`org.radarcns.avro.HeartRateItem` should become `heart_rate`
the current `heart_rate` should become `value`
`descriptive_statistic` and `unit` should be embedded in a new `header` value.
we have `beats_per_min` instead of `beats/min` since `/` cannot be used in Avro since it is a special character.

herkulano:
Changed JSON example:
https://github.com/RADAR-CNS/RADAR-Schemas/blob/1cc2d8705e4ce6661914fb404b387ad2ff38bc12/REST-API/Heart%20Rate/heart-rate.json

afolarin:
I'm doing a little digging about the OMH dataset or lack of it. Will report back.

herkulano:
From today's call, we decided that this is a good basis to start.

PR #16, opened 2017-03-30
Phone update

blootsvoets:
Added new phone sensors.

### DIFF
--- a/common/phone/phone_location.avsc
+++ b/common/phone/phone_location.avsc
@@ -1,0 +1,17 @@
+{
+  "namespace": "org.radarcns.phone",
+  "type": "record",
+  "name": "PhoneLocation",
+  "doc": "Data from the gps and network location providers.",
+  "fields": [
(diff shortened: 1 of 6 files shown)


common/phone/phone_location.avsc
```diff
@@ -0,0 +1,17 @@
+{
+  "namespace": "org.radarcns.phone",
+  "type": "record",
+  "name": "PhoneLocation",
+  "doc": "Data from the gps and network location providers. The latitude and longitude are stated with an unspecified reference offset and can thus be treated as relative locations. They cannot be used to infer absolute location. This means accurate distances or angles between locations cannot be calculated since those depend on the absolute location.",
```
afolarin: should we have the possibility of capturing both relative and absolute phone location? There are real senarios where one might have permission to gather this data

blootsvoets:
Good point, I'll rename this to PhoneRelativeLocation. We can add another if we need absolute locations as well.

PR #17, opened 2017-04-04
Publish the schemas as a Java API

blootsvoets:
Like [OpenMHealth does](https://github.com/openmhealth/schemas), we should really publish an SDK with the schemas. This makes versioning the commons library independent of the schemas used, which is nice if we don't need to change the commons library itself. This has the added benefit that we can use travis to test if the Avro files are valid.


blootsvoets:
As OpenMHealth does, I think it makes sense to allow our most common use case, using Java, available with the schemas, using the same versions of Java classes and Avro schemas. This way we don't have to keep versions in sync, and since its all generated code, I don't think it makes sense to create a separate repo for it. Plus, it actually checks if schemas are valid and will yield valid generated code at the time that the schemas are created.

fnobilia:
I really appreciate your idea of having a Java SDK, but one of the reasons for using AVRO is to have a useful way to serialise and deserialise complex objects on a language independent way. Some of these schemas will be used also by other projects that are not coded in JAVA like for instance the Questionnaire app or a possible native app for iOS. Furthermore we may need to create a library also for Python or for Swift, how we can support them then?

I would suggest to have the master branch as much generic as we can, and then using different branches to provide libraries, in this way we can cover multiple scenarios.

Is there any other solution to provide libraries coded in different programming languages relying on a single branch?

@afolarin what do you think?

blootsvoets:
Language independence of the Avro schemas is not at question here. They can always be used independently.

1. languages don't NEED this SDK, it just speeds up the development. For example, in Python or Javascript you wouldn't use an SDK, you'd just use the schema and it will dynamically create the objects. In Java you can in principle do the same (with GenericContainer), however, its far more convenient to provide the POJO classes.
2. We don't have any other languages to support right now. If we have another big language to support, such as swift, I think we should put that here, and again, the SDK should always match the schema.

I'm fine with letting the questionnaire out of this if they don't plan on using Java and we don't plan on supporting the schemas from a Java library. However, having this SDK for at least commons is just very very practical.

Alternatives:
1. creating again new repos that have to track this repo, this will bring inconsistencies.
2. excluding the java files from git history and just providing the build scripts so that we can build a binary to upload to bintray. This way the repo isn't directly useful to use as a submodule but at least the schemas are checked by travis and the binaries can be uploaded.

Branches are in my opinion meant for short-running alternative histories that can always be merged. We can set up an automatically tracking branch that copies schemas and creates an sdk from it, but it doesn't decrease the repo size, and it will obfuscate what the library does.

fnobilia:
> Language independence of the Avro schemas is not at question here. They can always be used independently.
Yes but adding `Gradle` scripts to the repository we end up with something that see quite java based. `Gradle` supports java, groovy and scala.
> We don't have any other languages to support right now. If we have another big language to support, such as swift, I think we should put that here, and again, the SDK should always match the schema.
Correct, but this is a long project we do not know what we could need to support in 5 years time, being generic should help to be flexible.

I'm not questioning the value of the SDK, I'm questioning the generalisation of the repository structure.
Could we move all the gradle files in a separate folder and then ask `Travis` to run the gradle scprits from that folder? In this way the root folder will be more generic and clean.

blootsvoets:
Fair point. I've moved all Gradle files to a `java-sdk` folder in #18. Closing this pull request.

PR #23, opened 2017-04-20
config/v1/config.json added.

kerzmaximilian:
@herkulano @petcupaula: Discussed with @fnobilia to add the configuration file to the schema repository. I also added a `../vX/..` to the route. I am assuming that layouts will change and maybe we want to support multiple versions. Let me know if that is to convoluted and I'll remove it from the path.

@fnobilia could you have a look at format and documentation?

### DIFF
--- a/restapi/config/v1/config.json
+++ b/restapi/config/v1/config.json
@@ -1,0 +1,172 @@
+{
+  "sensor": {
+    "ACCELEROMETER": {
+      "key": "acceleration",
+      "doc": "Acceleration item for a dataset with gravitational constant g as unit.",
+      "label": {
+        "EN": "Accelerometer"
+      },
+      "subKeys": [
+        {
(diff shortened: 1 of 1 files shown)


fnobilia:
@kerzmaximilian I suggested to store this schema here to have all `RADAR-CNS schemas` in the same repository.

kerzmaximilian:
@fnobilia yep - good point! It might be better to remove the versioning from the config file and apply it to the entire api down the line if there are major changes? (Eg.: `https://z.z.z.z:80/v1/...`)

fnobilia:
@kerzmaximilian what do you mean?

kerzmaximilian:
@fnobilia just quickly putting in writing what we discussed: generally agree to versioning of the api, but as you suggested, it is better to do this through versioned containers and nginx.

PR #25, opened 2017-04-28
Dashboard config

kerzmaximilian:
@herkulano could you have a quick review of the schema? If alright, Francesco will deploy it.


fnobilia:
Max, I would suggest to put it on folder different from `REST API`. Maybe dashboard, labels.

fnobilia:
@kerzmaximilian I'd suggest to move this configuration in `Firebase`, in this way all RADAR remote config files will be in a single place.

herkulano:
@kerzmaximilian agree with @fnobilia moving it to Firebase.

fnobilia:
Agreed on moving the configuration file on `Firebase`. I leave the branch opened, in this way the config file will be still accessible via `REST-Api`. I will remove the branch when the `Firebase` implementation will be finalised.

@herkulano @kerzmaximilian

PR #26, opened 2017-05-09
Time frame

fnobilia:
New version of Dataset.avsc.
Item.avsc reports only the `start` point in time. The `end` point can be computed using the `timeFrame` contained in the dataset's `header`. The `timeFrame` states the time interval between to consecutive samples.

The name `DoubleSample` is border line according to our guideline. It represents a generic sample represented by a double value.


restapi/dataset/item.avsc
```diff
@@ -4,15 +4,10 @@
...
-      "org.radarcns.avro.restapi.sensor.Acceleration" ,
-      "org.radarcns.avro.restapi.sensor.Battery" ,
-      "org.radarcns.avro.restapi.sensor.BloodVolumePulse" ,
-      "org.radarcns.avro.restapi.sensor.ElectroDermalActivity" ,
-      "org.radarcns.avro.restapi.sensor.InterBeatInterval" ,
-      "org.radarcns.avro.restapi.sensor.Temperature"
-      ] } ,
-    {"name": "effectiveTimeFrame", "type": "org.radarcns.avro.restapi.header.EffectiveTimeFrame"}
+    {"name": "sample", "type": [
+      "org.radarcns.avro.restapi.data.DoubleSample" ,
```
blootsvoets: Why use DoubleSample at all, and not just expand to `["double", "Quartile", "Acceleration"]`? I'm not sure that `sample` is a better description than `value`, since the rest API currently does not return samples but only statistics.

restapi/header/header.avsc
```diff
@@ -4,8 +4,9 @@
   "name": "Header",
   "doc": "Metadata for Dataset. It is a summary explaining what kind of information are stored in the Dataset.",
   "fields": [
-    {"name": "descriptiveStatistic", "type": "DescriptiveStatistic" },
-    {"name": "unit", "type": "org.radarcns.avro.restapi.sensor.Unit"},
-    {"name": "effectiveTimeFrame", "type": "EffectiveTimeFrame" }
+    {"name": "descriptiveStatistic", "type": "DescriptiveStatistic", "doc": "Statical value expressed by samples" },
```
blootsvoets: Would this also be a good place to indicate what type of data you're actually sending? Heart rate, acceleration?

restapi/dataset/item.avsc
fnobilia: The idea is that you first have a `data class` representing the aggregated `sample`, and then the value may be either a single `double` or a `Quartile`. The next iteration will be to add the `quality` field to `Acceleration` and `DoubleSample`.

restapi/header/header.avsc
fnobilia: The idea was that since the user needs to provide `userId`, `sourceId`, and `sensor` these 3 informations were redundant. While `descriptiveStatistic` is need to understand if the `value` of the returned sample is a `Quartile` or not.

blootsvoets: It's more that when a query and the answer are not directly connected, such as with a data download or backup, or you have a call that returns data from multiple datasets, it is still possible to disambiguate what the dataset is describing. Plus, it would only be one-time fields in the header rather than repeating values in the dataset items.

restapi/dataset/item.avsc
blootsvoets: Aha ok. Btw: I'm still not sure how to interpret this quality in an aggregated statistic. Suppose I have a descriptive statistic `MIN` over time period T and it has quality 0.6, is that also the minimum quality over that period T? Or would we send the same mean quality value over T for every descriptive statistic `MIN`, `MAX`, `SUM`, etc.?

fnobilia: Since all statical values are computed using the same values, all `DescriptiveStatistic` for period `T` will have the same quality.

BTW, there was also the idea of providing more information about the data quality running some analysis from `Restructure-HDFS-topic` while restructuring. For example we may compute the distribution of missing data: let's consider:

- a sensor generating 1 sample per minutes
- a situation where we have lost 60 samples
- hour dataset covers 1 day.

If these 60 lost messages are consecutive (meaning that we have 1h of gap), the quality of the dataset should be lesser than a dataset where we have lost 3 messages per hour every 20 min.

fnobilia:
@blootsvoets if everything is ok, I'm ready to merge and publish an `alpha` release

restapi/header/header.avsc
```diff
@@ -4,8 +4,13 @@
   "name": "Header",
   "doc": "Metadata for Dataset. It is a summary explaining what kind of information are stored in the Dataset.",
   "fields": [
-    {"name": "descriptiveStatistic", "type": "DescriptiveStatistic" },
-    {"name": "unit", "type": "org.radarcns.avro.restapi.sensor.Unit"},
-    {"name": "effectiveTimeFrame", "type": "EffectiveTimeFrame" }
+    {"name": "subjectId", "type": "string", "doc": "Subject identifier"},
+    {"name": "sourceId", "type": "string", "doc": "Source identifier"},
+    {"name": "source", "type": "org.radarcns.avro.restapi.source.SourceType", "doc": "Source type"},
+    {"name": "sensor", "type": "org.radarcns.avro.restapi.sensor.SensorType", "doc": "Sensor that has generated data"},
```
blootsvoets: Could this field and the next four also be `null`? That could perhaps also be the default? Is this header only for statistics, or also possibly for raw data (like phone status)?

fnobilia: To use this `header` with raw data, I would suggest to make `descriptiveStatistic` and `timeFrame` `nullable`. But to return raw data using `Dataset`, the `Item` schema has to support all raw data schemas and `startDateTime` can be `nullable`. Just to make it clearer
```json
{
  "namespace": "org.radarcns.avro.restapi.dataset",
  "type": "record",
  "name": "Item",
  "doc": "Dataset item. It contains a value and its related time.",
  "fields": [
    {"name": "sample", "type": [
      "org.radarcns.avro.restapi.data.DoubleSample" ,
      "org.radarcns.avro.restapi.data.Acceleration",
      "org.radarcns.empatica.EmpaticaE4Acceleration",
      "org.radarcns.biovotion.BiovotionVSMAcceleration",
      "..."
      ], "doc": "..." } ,
    {"name": "startDateTime", "type": ["null", "string"], "doc": "...", "default": null}
  ]
}
```

blootsvoets: Yes, next time is fine.

PR #34, opened 2017-07-07
Rename inpu_type.avsc to input_type.avsc

jorgecid:


fnobilia:
@jorgecid  I am going to close this pull request. Instead of changing the base branch as @blootsvoets  suggested. I would prefer if you can either directly update the schema in the `thinc-it` brach or make a pull request on the `thinc-it` branch. In this way the repository history will be cleaner.

Issue #45, opened 2017-07-28
Guideline for Default value


fnobilia:
We should define a guideline about `default` values.

In AVRO, `default` values are crucial to ensure schema evolution. Quoting the [Avro Specification 1.8.2](https://avro.apache.org/docs/current/spec.html)
> if the reader's record schema has a field that contains a default value, and writer's schema does not have a field with the same name, then the reader should use the default value from its field. If the reader's record schema has a field with no default value, and writer's schema does not have a field with the same name, an error is signalled.

In case of UNION with `null`, AVRO provides two approaches:
- `optional` values, the default value is `null` because the UNION is defined like `"type": ["null", "string"]`
- `nullable` value, the default value class must match the no-null schema. For instance in case `"type": ["string", "null"]` the default value must be a `string`
For union the default value must be an instance of the first type appearing in the UNION definition.

In case of ENUM, we can force a symbol `UNKNOWN` as default value.

In all other cases, the policy to apply is left to the developer.

Applying the correct policy can also simplify the downstream analysis.

fnobilia:
@mbegalevibrent waiting for your update

blootsvoets:
Another reason not to prefer `NaN` as a default double value: the `AvroData.fromConnectSchema` chokes on it:

```
org.apache.avro.SchemaBuilderException: org.codehaus.jackson.JsonParseException: Non-standard token 'NaN': enable JsonParser.Feature.ALLOW_NON_NUMERIC_NUMBERS to allow
  at [Source: java.io.StringReader@94a86f7; line: 1, column: 7]
  at org.apache.avro.SchemaBuilder.toJsonNode(SchemaBuilder.java:2595)
  at org.apache.avro.SchemaBuilder.access$4900(SchemaBuilder.java:217)
  at org.apache.avro.SchemaBuilder$FieldBuilder.completeField(SchemaBuilder.java:2115)
  at org.apache.avro.SchemaBuilder$FieldBuilder.access$5300(SchemaBuilder.java:2034)
  at org.apache.avro.SchemaBuilder$GenericDefault.withDefault(SchemaBuilder.java:2423)
  at io.confluent.connect.avro.AvroData.addAvroRecordField(AvroData.java:679)
  at io.confluent.connect.avro.AvroData.fromConnectSchema(AvroData.java:613)
```

blootsvoets:
Closed with #55

PR #49, opened 2017-08-30
No nan

blootsvoets:
Replace `NaN` with `null` as per #45. Also updated gradle.


commons/phone/phone_relative_location.avsc
```diff
@@ -6,12 +6,17 @@
...
+    {"name": "latitude", "type": ["null", "double"], "doc": "relative latitude from an initial random reference latitude (degrees with range [-90, 90])", "default": null},
+    {"name": "longitude", "type": ["null", "double"], "doc": "relative longitude from an initial reference longitude (degrees with range [-180, 180])", "default": null},
+    {"name": "altitude", "type": ["null", "float"], "doc": "height above the WGS 84 reference ellipsoid (m)", "default": null},
+    {"name": "accuracy", "type": ["null", "float"], "doc": "The estimated horizontal accuracy of this location, radial (m). We define horizontal accuracy as the radius of 68% confidence. In other words, if you draw a circle centered at this location's latitude and longitude, and with a radius equal to the accuracy, then there is a 68% probability that the true location is inside the circle.", "default": null},
+    {"name": "speed", "type": ["null", "float"], "doc": "speed over ground (m/s)", "default": null},
+    {"name": "bearing", "type": ["null", "float"], "doc": "the horizontal direction of travel of this device (degrees with range (0, 360])", "default": null}
```
fnobilia: We may follow this approach on all other schemas: `time` and `timeReceived` cannot be null and do not have default, while all the other fields can be either `optional`  fields or enum with `UNKNOW` as default.

@blootsvoets What do you think?

blootsvoets: Yes, that would be my preferred approach for schemas with a lot of values that are may or may not be set. For some schemas, having `null` defaults in a field would add less value. It does slightly increase the processing needed (union traversal and intermediate objects) and additional checks at the analysis stage (remove/correct null values before actual analysis). Once a field has union-null type, a null check is always necessary.

PR #52, opened 2017-09-01
Added a scheme for phone Bluetooth devices info

strangepleasures:


commons/phone/phone_bluetooth_devices.avsc
```diff
@@ -0,0 +1,13 @@
+{
+  "namespace": "org.radarcns.phone",
+  "type": "record",
+  "name": "PhoneBluetoothDevices",
+  "doc": "Phone Bluetooth devices info.",
+  "fields": [
+    {"name": "time", "type": "double", "doc": "device timestamp in UTC (s)"},
+    {"name": "timeReceived", "type": "double", "doc": "device receiver timestamp in UTC (s)"},
+    {"name": "pairedDevices", "type": "int", "doc": "Number of paired devices"},
+    {"name": "nearbyDevices", "type": "int", "doc": "Number of nearby devices"},
```
blootsvoets: perhaps make the type `["null", "int"]`, so it can be `null` whenever bluetooth is not enabled. Same for `pairedDevices`.

commons/phone/phone_bluetooth_devices.avsc
```diff
@@ -0,0 +1,13 @@
...
+    {"name": "enabled", "type": "boolean", "doc": "whether Bluetooth is enabled"}
```
blootsvoets: perhaps change the name to `bluetoothEnabled`?

commons/phone/phone_bluetooth_devices.avsc
strangepleasures: Actually the plugin always turns Bluetooth on (and switches it off back if it wasn't active), so it should be always able to get the numbers.

commons/phone/phone_bluetooth_devices.avsc
strangepleasures: `bluetoothEnabled` would be indeed a better name, but you have already merged RADAR-Android-Phone branch which uses `enabled`. Can we maybe go with `enabled`, otherwise a new merge request for RADAR-Android-Phone would be needed?

strangepleasures: Done

commons/phone/phone_bluetooth_devices.avsc
```diff
@@ -0,0 +1,13 @@
+{
+  "namespace": "org.radarcns.phone",
+  "type": "record",
+  "name": "PhoneBluetoothDevices",
+  "doc": "Phone Bluetooth devices info.",
+  "fields": [
+    {"name": "time", "type": "double", "doc": "device timestamp in UTC (s)"},
+    {"name": "timeReceived", "type": "double", "doc": "device receiver timestamp in UTC (s)"},
+    {"name": "pairedDevices", "type": ["null", "int"], "doc": "Number of paired devices"},
+    {"name": "nearbyDevices", "type": "int", "doc": "Number of nearby devices"},
```
blootsvoets: Could you set this type to `["null", "int"]` as well, and I just realised, for both the default can also be set to `"default": null`.

strangepleasures: Done

commons/phone/phone_bluetooth_devices.avsc
strangepleasures: Done

Issue #40, opened 2017-07-10
Folder layout


fnobilia:
After a discussion with @blootsvoets, we have identified the need to refactor the schema repository structure.

Instead of having a single `commons` root folder for all common schema, we can specialise it in `passive`, `active` and `kafka`.

`passive` folder will contain all schemas used by the passive sources. These schemas must have
 - `time` as first field, and it represents the `timestamp in UTC (s)` attached by the wearable device to the sample formatted in `double`
 - `timeReceived ` as second field, it represents the `timestamp in UTC (s)` attached by the smartphone when it receive the sample from the wearable device. It must be formatted in `double`

`active` folder will contain all schemas used by the active sources. These schemas must have
 - `time` as first field, formatted as `double`, it represents the `timestamp in UTC (s)` when the assessment has been submitted to the subject 
 - `timeCompleted ` as second field, , formatted as `double`, it represents the `timestamp in UTC (s)` when the subject has completed the assessment

`kafka` folder will contain all schemas internally used in Kafka.

The final tree folder structure will be:

- commons
  * active
  * kafka
  * passive
- rest

Having a common `time` field will simplify the logic in the `HDFS-Extractor` as well as it can provide a mechanism to compute temporal order across all collected data.

Unit tests should be provided to make sure schemas in each specific folder matches the requirements.

`README.md` should be updated to state the suggested schema layouts.

blootsvoets:
An addition to the update and README: each folder for a passive or active source type should contain a `README.md` with links to the relevant hardware, specifications, developer information and RADAR implementation.

fnobilia:
The folder tree can be augmented with a `specification` folder, again split in `passive` and `active`.

- commons
  * active
  * kafka
  * passive
- rest
- specification
  * active
  * passive

`specification/active` will contain `yml` files defining assessments. For example, the `phq_8.yml` file will state questions, possible answers and `RADAR tool` used to submit the question to the subject. 

`specification/passive` will contain `yml` files defining device specifications. The specification will list all on-board sensor. For each sensor there will be
- name
- sample rate
- unit
-`data_type` (define the sample is RAW data or it is the result of a vendor computation or a RADAR computation).

Other possible information at sensor level can be: kafka topic names and avro schemas.

A suitable starting point for the `active specification layout` is [here](https://github.com/RADAR-CNS/RADAR-Schemas/blob/phq8/questionnaire/specification/PHQ8_config.json).

[...]

Issue #28, opened 2017-05-22
Topic list programmatically generated


fnobilia:
If all the device projects expose topic names as constants, then another simple java solution having all the device projects as dependencies can programmatically generate an output file used by the `install-radar-stack.sh` to update `.env`.

From a discussion with @blootsvoets 

> those jar could have a single command that simply lists topics, with flags like `--raw`, `--processed`, `--schemaless` to only list those topics so we can use them in different commands and `-0` to have them `'\0'`-separated. That could be a jar file, runnable like
```shell
$ java -jar schemas.jar topics --raw
android_empatica_e4_acceleration
android_empatica_e4_battery_level
android_empatica_e4_blood_volume_pulse
...
```

We may expose also a flag for each device, something like `--empatica` `--smartphone-sensor`.

Another idea is to provide more metadata in output. For instance, reusing `org.radarcns.avro.restapi.sensor.SensorType` a possible result may be
```shell
$ java -jar schemas.jar topics --raw
ACCELEROMETER: android_empatica_e4_acceleration
BATTERY: android_empatica_e4_battery_level
BLOOD_VOLUME_PULSE: android_empatica_e4_blood_volume_pulse
...
```
Then adding the flag `-q`, it will return only the topic list.
```shell
$ java -jar schemas.jar topics --raw -q
android_empatica_e4_acceleration
android_empatica_e4_battery_level
android_empatica_e4_blood_volume_pulse
...
```

blootsvoets:
Done with #55

PR #55, opened 2017-09-21
Restructuring

blootsvoets:
This continues the huge effort started by Francesco to get a more organised repository. It consists of the following changes:
- The schemas represented are now organised as discussed as #40. This is a breaking change, as it changes the namespace and some names of schemas. To use the new RADAR-Schemas repository, the Kafka server and Hadoop data storage should be cleared and reset.
- There is a `radar-schema` binary that 
  - validates the defaults as discussed in #45
  - validates the time fields
  - validates the documentation of schemas
  - lists the topic names as discussed in #28

The new `specifications` directory contains the data needed to list the topics and show some metadata about it. It also contains the data of questionnaires.


blootsvoets:
@afolarin @yatharthranjan @strangepleasures @mbegalevibrent are you okay with these changes? I've got a thumbs-up, but I'd like to be sure before we go ahead with this.

java-sdk/radar-schemas-tools/README.md
```diff
@@ -0,0 +1,92 @@
...
+- commons
+  * active
+  * kafka
+  * monitor
+  * passive
+- rest
+- specification
+  * active
+  * monitor
+  * passive
```
afolarin: add:
specification
* stream

java-sdk/radar-schemas-tools/README.md
```diff
@@ -0,0 +1,92 @@
+# RADAR Schemas Validator
+
+The RADAR Schemas Validator checks if the `Schema Catalog` is in a valid state.
+
+It first checks the folder structure, it has to be compliant with:
+- commons
+  * active
+  * kafka
+  * monitor
+  * passive
```
afolarin: add 
commons
* stream
* catalogue

java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/SchemaRepository.java
```diff
@@ -0,0 +1,51 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.radarcns.schema;
+
+/**
+ * Checks the schema catalog.
+ * Checks that the tree folder structure respects the following structure:
```
afolarin: again perhaps needs updating to reflect latest restructuring folder structure

java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/specification/active/questionnaire/QuestionnaireDataTopic.java
```diff
@@ -0,0 +1,48 @@
...
+import javax.validation.constraints.NotEmpty;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * TODO.
+ */
+public class QuestionnaireDataTopic extends DataTopic {
+    public enum RadarSourceTypes {
+        PHQ8
```
afolarin: PHQ8 enum: should these types be derived from the specification catalogue TYPE (otherwise do we have to define here again)?

java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/specification/passive/PassiveDataTopic.java
```diff
@@ -0,0 +1,60 @@
...
+import org.radarcns.catalogue.ProcessingState;
+import org.radarcns.schema.specification.AppDataTopic;
+
+import java.util.Objects;
+
+/**
+ * TODO.
+ */
+public class PassiveDataTopic extends AppDataTopic {
+    public enum RadarSourceDataTypes {
```
afolarin: again similar to PHQ8 enum: should these types be derived from the specification catalogue TYPE (otherwise do we have to define here again)?

java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/specification/active/questionnaire/QuestionnaireDataTopic.java
blootsvoets: For extensibility, the specifications use a `string` to define the type of questionnaire or data source. That way, someone can add specifications with types not present in RADAR. For validation, we do check if the given types are valid within RADAR. Concretely:

- Within RADAR-CNS, we presently only support the PHQ8 questionnaire. For any other value, the validator will notify the user that this is not a RADAR questionnaire. If another type is needed, it should be added to this enum.
- Outside RADAR-CNS, for example Thinc-It may want to add a custom `THINCIT-Q1` type. If the specification itself was an enum, this would not parse correctly. Now it will parse, but it will just not be recognised as a RADAR questionnaire.

afolarin: ok clear now.

PR #64, opened 2017-11-15
Add catalogue service

nivemaham:
Added a server to share source-catalogues through web services
Updated dev with master.


java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/service/SourceCatalogueServer.java
```diff
@@ -0,0 +1,89 @@
...
+
+    public static SubCommand command() {
+        return new SourceCatalogueServiceCommand();
+    }
+
+    private static class SourceCatalogueServiceCommand implements SubCommand {
+
+        @Override
+        public String getName() {
+            return "catalogue-server";
```
blootsvoets: would just `serve` would be enough?

java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/service/SourceCatalogueServer.java
```diff
@@ -0,0 +1,89 @@
...
+                return 1;
+            }
+            service.close();
+            return 0;
+        }
+
+        @Override
+        public void addParser(ArgumentParser parser) {
+            parser.description("Create all topics that are missing on the Kafka server.");
+            parser.addArgument("-port")
```
blootsvoets: for the other arguments, POSIX flags are used, so `"-p", "--port"`.

README.md
```diff
@@ -58,4 +58,7 @@ docker-compose run --rm tools radar-schemas-tools list
 docker-compose run --rm tools radar-schemas-tools register http://schema-registry:8081
 # create topics with zookeeper
 docker-compose run --rm tools radar-schemas-tools create zookeeper-1:2181
+# run source-catalogue webservice
+docker-compose run --rm tools radar-schemas-tools catalogue-server (optional-port <portnumber>)
```
blootsvoets: perhaps `serve` would be enough? The port would be `-p <portnumber>`

PR #67, opened 2017-11-17
Adding properties to support source-type import in MP

nivemaham:
Small correction

### DIFF
--- a/java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/service/SourceCatalogueServer.java
+++ b/java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/service/SourceCatalogueServer.java
@@ -77,7 +77,7 @@ public class SourceCatalogueServer implements Closeable {
         @Override
         public void addParser(ArgumentParser parser) {
             parser.description("Create all topics that are missing on the Kafka server.");
-            parser.addArgument("-p" ,"-port")
+            parser.addArgument("-p" ,"--port")
                     .help("Port number of the SourceCatalogue Server ")
                     .type(Integer.class)
                     .setDefault(9010);
--- a/java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/service/SourceCatalogueServer.java
+++ b/java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/service/SourceCatalogueServer.java
@@ -76,7 +76,7 @@ public class SourceCatalogueServer implements Closeable {
 
         @Override
         public void addParser(ArgumentParser parser) {
-            parser.description("Create all topics that are missing on the Kafka server.");
+            parser.description("Port number of the Catalogue server.");
             parser.addArgument("-p" ,"--port")
                     .help("Port number of the SourceCatalogue Server ")
                     .type(Integer.class)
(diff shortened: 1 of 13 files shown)


java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/service/SourceCatalogueServer.java
```diff
@@ -77,7 +77,7 @@ public int execute(Namespace options, CommandLineApp app) {
         @Override
         public void addParser(ArgumentParser parser) {
             parser.description("Create all topics that are missing on the Kafka server.");
```
blootsvoets: Also missed in the previous review: please update description

java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/service/SourceCatalogueServer.java
```diff
@@ -76,8 +76,8 @@ public int execute(Namespace options, CommandLineApp app) {
 
         @Override
         public void addParser(ArgumentParser parser) {
-            parser.description("Create all topics that are missing on the Kafka server.");
-            parser.addArgument("-p" ,"-port")
+            parser.description("Port number of the Catalogue server.");
```
blootsvoets: This is the description of the task (I.e. web service).

specifications/active/thincit.yml
```diff
@@ -1,4 +1,7 @@
 name: THINCIT
+vendor: LUNDBECK
```
blootsvoets: They are just thincit iirc.

specifications/active/thincit.yml
```diff
@@ -1,4 +1,7 @@
...
+model: THINCIT
+version: v1
```
blootsvoets: semantic versioning would be preferable. This could be `1.0.0`

java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/specification/DataProducer.java
```diff
@@ -26,9 +26,6 @@
     @JsonProperty
     private List<String> labels;
 
-    @JsonProperty
-    private String version;
```
blootsvoets: Why not add the other two properties here, instead of adding them to the subclasses?

java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/service/SourceCatalogueServer.java
nivemaham: Indeed. My bad. Will correct

java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/specification/DataProducer.java
nivemaham: The reason i moved it to AppSource is having those in the `DataProducer` also affects `StreamGroups`. I am not sure whether we want that?. @blootsvoets

specifications/active/thincit.yml
nivemaham: Good idea. For all source-types i assume. Will change them

PR #70, opened 2017-11-27
prepare release

nivemaham:
Is there anything else we should do to prepare a new release of RADAR-Schemas? @blootsvoets

### DIFF
--- a/java-sdk/build.gradle
+++ b/java-sdk/build.gradle
@@ -17,7 +17,7 @@ subprojects {
     apply plugin: 'idea'
 
     // Configuration
-    version = '0.2.1-SNAPSHOT'
+    version = '0.2.1'
     group = 'org.radarcns'
     ext.githubRepoName = 'RADAR-CNS/RADAR-Schemas'
 


yatharthranjan:
Hi @nivemaham, We need to add schemas and specifications for the aRMT app for all the questionnaires. @kerzmaximilian will need them for the release of aRMT before this wednesday

nivemaham:
@yatharthranjan , I am waiting for the confirmation of @blootsvoets for the release. Do you need the released distribution? i do not expect changes in the specifications or schemas. We should be able to do the release tomorrow, i believe.

yatharthranjan:
If we add the schemas and specification to the repo, will these be automatically registered without them in the release?

blootsvoets:
@yatharthranjan we can make a new release when the aRMT schemas get added. They will not be automatically be used until they are released _and_ the `kafka-radarinit` image is updated with the new release version. In RADAR-Docker, you can add specifications and schemas which are not yet released by adding them to the `commons` and `specifications` directories in `etc/schema`.

PR #72, opened 2017-11-28
Release 0.2.1 

nivemaham:
Introduces `catalog server` to share `source-type specifications` provided here. This introduces new properties to uniquely identify a specification version by the combination of `vendor`, `model`, `version`. These three properties are required to create a source-type on ManagementPortal of RADAR. 

The catalog server can be run from `radar-schema-tools` with `serve -p <port>`


nivemaham:
This PR is closed and will be merged from intermediate branch following the usual practice.

PR #75, opened 2017-11-29
version specific specs

nivemaham:
Added version-ed specifications to support client-apps.


blootsvoets:
Note: there has to be a change in the Java validator. It expects PHQ8 to always be there, but that was replaced by aRMT.

PR #81, opened 2017-11-30
Add armt thincit schemas

### LINKED ISSUE #71: Add all schemas and topics for active RMT app and thinc it
Currently only one questionnaire topic and schema is added. We need to add the topics and schemas for all the topics

yatharthranjan:
closes #71


commons/active/thincit/thinc_it_pdq.avsc
```diff
@@ -0,0 +1,29 @@
+{
+  "namespace": "org.radarcns.active.thincit",
+  "type": "record",
+  "name": "ThincItPdq",
+  "doc": "THINC-it Assessment derived from PDQ5.",
+  "fields": [
+    { "name": "time", "type": "double", "doc": "Timestamp in UTC (s) when the test is submitted to the subject." },
+    { "name": "timeCompleted", "type": "double", "doc": "Timestamp in UTC (s) when the subject completes the test." },
+    { "name": "score", "type": "int", "doc": "THINC-it index score - 0 to 4000." },
+    { "name": "questionOne", "type":
```
blootsvoets: Is there a possibility that not all questions are answered? In that case, the questions should probably be made type nullable `["null", {"type": "enum" ...}]`, with `"default": null`.

yatharthranjan: I don't think it will be null. It only sends info after everything is complete. I will update it in any case just to be safe.

PR #86, opened 2018-01-09
Added aggregate list and name to double aggregate

blootsvoets:
This way, aggregates can be identified and there is a general list. This eases conversion to MongoDB and subsequently the REST API.


commons/stream/aggregator/aggregate_list.avsc
```diff
@@ -0,0 +1,10 @@
+{
+  "namespace": "org.radarcns.stream.aggregator",
+  "type": "record",
+  "name": "AggregateList",
+  "doc": "Result of data aggregation.",
```
nivemaham: Maybe a more specific doc for this would help to have a clear idea why we moved to this schema.

PR #91, opened 2018-01-16
Source statistics

blootsvoets:
Make separate SourceStatistics class to monitor statistics about sources in the Backend.


java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/specification/stream/StreamDataTopic.java
```diff
@@ -27,12 +25,14 @@
     private Unit unit;
 
     @JsonProperty("input_topic")
-    @NotBlank
     private String inputTopic;
 
     @JsonProperty("topic_base")
```
nivemaham: LGTM. Just curious. What is the `topic_base` ? Main topics devices stream to and  streams consume from ?

blootsvoets: `input_topic` could be `android_empatica_e4_inter_beat_interval` and `topic_base` could be `android_empatica_e4_heart_rate`, with `windowed: true`. Then the output topics would become `android_empatica_e4_heart_rate_10sec`, `android_empatica_e4_heart_rate_10min`, etc.

nivemaham: Thought so.

PR #101, opened 2018-02-21
add schema and spec for active audio

yatharthranjan:


commons/active/opensmile/open_smile_2_audio_analysis.avsc
```diff
@@ -6,6 +6,8 @@
   "fields": [
     {"name": "time", "type": "double", "doc": "Device timestamp in UTC (s)."},
     { "name": "timeCompleted", "type": "double", "doc": "Timestamp in UTC (s) when subject completes the audio questionnaire." },
-    {"name": "data", "type": ["null", "string"], "doc": "Audio features computed by openSMILE recorded in Base64 format.", "default": null}
+    {"name": "config", "type": ["null", "string"], "doc": "OpenSMILE configuration file.", "default": null},
```
blootsvoets: Perhaps explicitly say config file contents? Or config file identifier, where the questionnaire app makes it completely clear what the possible identifiers are?

commons/active/opensmile/open_smile_2_audio_analysis.avsc
```diff
@@ -6,6 +6,8 @@
...
+    {"name": "format", "type": ["null", "string"], "doc": "Format of the recorded data.", "default": null},
```
blootsvoets: Is this audio format? If this is an analysis, will it have a format? What available formats are there? E.g. link to Android audio output file formats, or state that this means for example csv file.

blootsvoets:
@kerzmaximilian what is raw audio? Is that a WAV format? Then that is the format that should be stated in format. Also, if this is meant to just contain WAV audio, perhaps rename it to opensmile2audio instead and state so in the documentation. Unless it is unambiguously clear what data is contained in these records, automatic processing is almost impossible to do. Note that by using the Android media recorder with mpeg compression, recording, file transfers and data storage will be much more efficient.

blootsvoets:
There might be another way to unambiguously state the contents of the audio, just by listing the mime type. For 3GPP this is `audio/3gpp`, for MPEG-4 this is `audio/mp4`, for WAV it is `audio/wav`.

Btw, there is a standard [Cordova plugin for audio recording](https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-media-capture/#capturecaptureaudio) as well. It records the file with content and metadata, which includes a MIME-type:
> type: The file's mime type (DOMString)

yatharthranjan:
Hi Joris, the audio format could be for example - wav, mpeg, etc. Also `Cordova plugin for audio recording` cannot be used since it opens up a new interface which we do not want. We need the users to read the text on the aRMT app screen. Should I just add the types as example in the docs for format field?

yatharthranjan:
Continued in #104

PR #104, opened 2018-02-22
Audio active suggestion

blootsvoets:
Updated suggestion to #101, would this be a workable solution for you @kerzmaximilian @yatharthranjan?


yatharthranjan:
When will the 'questionnaire_audio_recording.avsc' be used ?? The aRMT will always encode the recorded data using opensmile and then post it to kafka. Otherwise looks good to me

blootsvoets:
@yatharthranjan How about this update, with `OpenSmile2AudioRecording`? I've added a `reciteText` property, or is that always the same?

yatharthranjan:
Looks good to me. What about you @kerzmaximilian ?

PR #111, opened 2018-03-01
updated repository to include fitbit schemas

mbegalevibrent:
activity steps, heart rate and sleep from fitbit API


mbegalevibrent:
@yatharthranjan -- confirming you see these!

commons/passive/fitbit/fitbit_activity_steps.avsc
```diff
@@ -0,0 +1,29 @@
...
+            "doc": "Type of data coming from fitbit intraday API, (e.g. heart rate)"
+        },{
+            "name": "datasetInterval",
+            "type": "int",
+            "doc":  "Chronological window size (e.g. 1)"
+        },{
+            "name": "datasetType",
+            "type": "string",
+            "doc": "Chronological window unit (e.g. 'minute')"
+        },{
```
blootsvoets: Could `datasetInterval` and `datasetType` be merged to a `interval` or `window` field, type `double`, containing `Chronological window size (s).`?

commons/passive/fitbit/fitbit_heart_rate.avsc
```diff
@@ -0,0 +1,30 @@
...
+            "type": "string",
+            "doc": "Chronological window unit (e.g. 'minute')"
+        },{
+            "name": "time",
+            "type": "string",
+            "doc": "Hours, minutes, seconds of start of interval (e.g. 00:33:00)"
+        },{
+            "name": "value",
+            "type": "int",
+            "doc":"Heart rate BPM value"
```
blootsvoets: Could be `Heart rate (bpm).`

commons/passive/fitbit/fitbit_sleep_classic.avsc
```diff
@@ -0,0 +1,25 @@
...
+        },{
+            "name": "level",
+            "type": "string",
+            "doc": "Level of sleep, 'awake', 'restless', or 'asleep'"
+        },{
+            "name": "dateTime",
+            "type": "string",
+            "doc": "Hours, minutes, seconds of start of interval"
+        },{
+            "name": "seconds",
```
blootsvoets: would something like `sleepDuration` be more informative?

commons/passive/fitbit/fitbit_sleep_classic.avsc
```diff
@@ -0,0 +1,25 @@
...
+  "doc": "Classic sleep data as defined at https://dev.fitbit.com/build/reference/web-api/sleep/",
+  "fields": [
+        {
+            "name": "type",
+            "type": "string",
+            "doc": "Type of data coming from fitbit intraday API, (e.g. sleep classic)"
+        },{
+            "name": "level",
+            "type": "string",
+            "doc": "Level of sleep, 'awake', 'restless', or 'asleep'"
```
blootsvoets: If this is not expected to change, I'd propose to make this an enum instead
```json
"type": {
   "name": "SleepStageLevel",
   "type": "enum",
   "symbols": ["DEEP", "LIGHT", "REM", "AWAKE", "RESTLESS", "ASLEEP", "UNKNOWN"],
   "doc": "Sleep level."
},
"doc": "Level of sleep.",
"default": "UNKNOWN"
```
I don't know how `restless` compares to `light` sleep, are there duplicated definitions in this symbol list?

commons/passive/fitbit/fitbit_sleep_stages.avsc
```diff
@@ -0,0 +1,25 @@
+{
+  "namespace": "org.radarcns.passive.fitbit",
+  "type": "record",
+  "name": "FitbitSleepStages",
```
blootsvoets: The fields of this schema seem identical to the ones in the `FitbitSleepClassic` schema. Is that intentional? Could we just use one schema `FitbitSleep`, and send the data to two topics (`fitbit_sleep_classic` and `fitbit_sleep_stages`)?

commons/passive/fitbit/fitbit_activity_steps.avsc
```diff
@@ -0,0 +1,29 @@
...
+            "name": "time",
+            "type": "string",
+            "doc": "Hours, minutes, seconds of start of interval (e.g. 00:33:00)"
+        },{
+            "name": "value",
+            "type": "int",
+            "doc":"Steps taken in this meriod"
```
blootsvoets: Is this steps taken? If so, could you please rename the field to `stepCount` or `steps`?

commons/passive/fitbit/fitbit_activity_steps.avsc
```diff
@@ -0,0 +1,29 @@
+{
+  "namespace": "org.radarcns.passive.fitbit",
+  "type": "record",
+  "name": "FitbitIntradayHeartRate",
+  "doc": "Intra day heart rate data from fitbit device",
+    "fields": [
+        {
+            "name": "type",
+            "type": "string",
+            "doc": "Type of data coming from fitbit intraday API, (e.g. heart rate)"
```
blootsvoets: Is this always steps? In that case, this field can be removed. Additional schemas could be made to accommodate other data types. If this schema is meant to be more general, perhaps rename it to something like `FitbitIntradayMetric` instead. Even then, perhaps the data can differ based on the topic it was sent to, e.g. `fitbit_intraday_heart_rate` or `fitbit_intraday_steps`, and the type field would not be necessary.

commons/passive/fitbit/fitbit_activity_steps.avsc
```diff
@@ -0,0 +1,29 @@
+{
+  "namespace": "org.radarcns.passive.fitbit",
+  "type": "record",
+  "name": "FitbitIntradayHeartRate",
```
blootsvoets: Rename to `FitbitActivitySteps`

commons/passive/fitbit/fitbit_heart_rate.avsc
```diff
@@ -0,0 +1,30 @@
+{
+  "namespace": "org.radarcns.passive.fitbit",
+  "type": "record",
+  "name": "FitbitIntradayHeartRate",
+  "doc": "Intra day heart rate data from fitbit device",
+    "fields": [
+        {
+            "name": "type",
+            "type": "string",
+            "doc": "Type of data coming from fitbit intraday API, (e.g. heart rate)"
```
blootsvoets: Is this always heart rate? In that case, this field can be removed. Additional schemas could be made to accommodate other data types. If this schema is meant to be more general, perhaps rename it to something like `FitbitIntradayMetric` instead. Even then, perhaps the data can differ based on the topic it was sent to, e.g. `fitbit_intraday_heart_rate` or `fitbit_intraday_steps`, and the type field would not be necessary.

commons/passive/fitbit/fitbit_heart_rate.avsc
```diff
@@ -0,0 +1,30 @@
...
+        },
+        {
+            "name": "datasetInterval",
```
blootsvoets: Could datasetInterval and datasetType be merged to a `interval` or `window` field, type `double`, containing `Chronological window size (s).`?

blootsvoets:
I see that the schema guidelines are not all in the same place. I'll move them from https://github.com/RADAR-base/RADAR-Schemas/tree/master/java-sdk/radar-schemas-tools to the main `README`.

yatharthranjan:
@mbegalevibrent I agree with @blootsvoets that wherever possible to use a generic schema, please just use one file for schema like - 

```avsc
/connector/fitbit/fitbit_intraday_generic.avsc
{
  "namespace": "org.radarcns.passive.fitbit",
  "type": "record",
  "name": "FitbitIntradayGeneric",
  "doc": "Intra day data from fitbit device",
    "fields": [
        {
            "name": "type",
            "type": "string",
            "doc": "Type of data coming from fitbit intraday API, (e.g. heart rate)"
        },
        {
            "name": "datasetInterval",
            "type": "int",
            "doc":  "Chronological window size (e.g. 1)"
        },{
            "name": "datasetType",
            "type": "string",
            "doc": "Chronological window unit (e.g. 'minute')"
        },{
            "name": "time",
            "type": "string",
            "doc": "Hours, minutes, seconds of start of interval (e.g. 00:33:00)"
        },{
            "name": "value",
            "type": "int",
            "doc":"Data value"
        }
  ]
}
```

And then in the specifications you can just do something like -

```
fitbit-1.0.0.yml

name: fitbit
vendor: fitbit
model: fitbit-connector
version: 1.0.0
data:
  - type: HEARTRATE
    topic: fitbit_intraday_heart_rate
    value_schema: .connector.fitbit.FitbitIntradayGeneric
  - type: STEPS
    topic: fitbit_intraday_steps
    value_schema: .connector.fitbit.FitbitIntradayGeneric


```

commons/passive/fitbit/fitbit_sleep_stages.avsc
yatharthranjan: Yes I agree. There is no need for two schemas

commons/passive/fitbit/fitbit_sleep_classic.avsc
yatharthranjan: Yes this is intentional since the fitbit API provides 2 measures of sleep based on processing data in 30 seconds interval and 60 seconds interval

yatharthranjan: Quoting the Fitbit WEB API 
```
This endpoint supports two kinds of sleep data:

stages : Levels data is returned with 30-second granularity. 'Sleep Stages' levels include deep, light, rem, and wake.

classic : Levels data returned with 60-second granularity. 'Sleep Pattern' levels include asleep, restless, and awake.
```

blootsvoets: Perhaps, this one can be named `FitbitSleepPattern` then, and the other still `FitbitSleepStage`?

blootsvoets:
@yatharthranjan On the contrary, I'd prefer not to merge the intraday measures. The contents of data are much more meaningful if the pertaining metadata and naming is added to the schema, rather than creating a generic container that does not really describe the data. The alternative is to have a generic schema and then needing to add a lot of metadata fields to describe the actual contents of the record.

yatharthranjan:
@blootsvoets Hmm. I don't see a lot of metadata right now. But if it is going to be something a lot like the Questions definitions in aRMT then definitely using different schemas makes more sense.

blootsvoets:
Metadata for me would be documentation on the unit used, the type of data collected. Having a column name `heartRate` instead of `value` seems useful to me, and extensible in case Fitbit adds more information to heart rate specifically.

yatharthranjan:
Okay agreed. In order to acommodate future changes best to have the separate schemas. 
BTW, **right now** the intraday data is basically a timeseries and contains a format like - 
```
    "activities-heart-intraday": {
        "dataset": [
            {
                "time": "00:00:00",
                "value": 64
            },
            {
                "time": "00:00:10",
                "value": 63
            },
            {
                "time": "00:00:20",
                "value": 64
            },
            {
                "time": "00:00:30",
                "value": 65
            },
            {
                "time": "00:00:45",
                "value": 65
            }
        ],
        "datasetInterval": 1,
        "datasetType": "second"
    }
```

and same for steps - 

```
"activities-steps-intraday":{
        "dataset":[
            {"time":"00:00:00","value":0},
            {"time":"00:01:00","value":0},
            {"time":"00:02:00","value":0},
            {"time":"00:03:00","value":0},
            {"time":"00:04:00","value":0},
            {"time":"00:05:00","value":287},
            {"time":"00:06:00","value":287},
            {"time":"00:07:00","value":287},
            {"time":"00:08:00","value":287},
            {"time":"00:09:00","value":287},
            {"time":"00:10:00","value":0},
            {"time":"00:11:00","value":0},
        ],
        "datasetInterval":1,
        "datasetType": "minute"
    }
```

So thats why I thought that just the topic name should be enough to distinguish.

mbegalevibrent:
I appreciate the spirit of these updates--the issue is that the UTC of the participant profile is not necessarily the UTC of the collected intraday data from a historical perspective.

"time" as a field cannot be guaranteed from a UTC perspective per the fitbit API--the data itself doesn't include information about timezone.

This works for prospective data, not retrospective data.

blootsvoets:
@mbegalevibrent do you mean that the timezone in the participant profile is not updated? The zone ID in their example at least does represent a DST-independent zone, e.g. `America/Los_Angeles`, meaning that for a given date-time and zone ID, there is a unique calculation to UTC. Or would this be a problem particularly when a participant travels?

mbegalevibrent:
Whether a participant travels or if they are near time zone borders or if their timezone is locked and they travel--more importantly, it's just that we can't know anything other than the relative time to the participant's device, not in UTC.  Does that make sense?

blootsvoets:
After a bit of searching, indeed this is an [open "bug"/missing feature](https://community.fitbit.com/t5/Web-API-Development/Clarification-on-time-zones-and-date-time-values/td-p/925255) in the Fitbit API. However, if we want to do any comparison between the Fitbit data and other data, we need the data in UTC. Note that the people who made the bug report to Fitbit are from OpenMHealth, one of the RADAR schemas inspirations.

Proposal:
- keep the time field `time`, and by mapping time stamps + configured zone ID to UTC. Rationale: it gives us a heuristic to sort most of the records in the correct bins (i.e. we assume that the timezone gives us the correct data in most cases). This logic is used in Kafka Streams, the MongoDB connector, REST API and HDFS connector.
- use `timezoneOffset` in any analyses that require the date-time reported by Fitbit. Regardless of whether the value for `time` was the correct UTC timestamp, `time + timzoneoffset` is equal to the date-time reported by Fitbit, it is a one-to-one mapping.

mbegalevibrent:
What about some slight modifications:

- keep the time field time, containing the mapping of local date time + configured zone ID to UTC. 
Rationale understood: In the schema documentation we need to ensure that this UTC isn't "valid" but based on an artificial calculation. I propose that we update this based on the expiration of the OAuth tokens (i.e. check for any updates and make it a "Best Guess"), so th

- use timezoneOffset in any analyses that require the date-time reported by Fitbit. Regardless of whether the value for time was the correct UTC timestamp, time + timzoneoffset is equal to the date-time reported by Fitbit, it is a one-to-one mapping.
Counter proposal:
 - Add a profile schema and polling route that gives us daily awareness of the timezone
 - Update connector to prefer push notifications

- we document the fact that the Fitbit API does not provide reliable time mappings, i.e., that the value of time may be incorrect but we have no way of knowing for which entries that is the case. Note that if location data is also tracked, you could choose to disregard data that is gathered far from the original source.
File a new bug report or append to the above one to note that we also require this feature.

blootsvoets:
That would be analogous to the pRMT phone streams as proposed, with all topics publishing to time, and a single application status topic mapping to timezone. My only preference for including the timezone offset in these records would be that combined with time, you have at least one reliable time point per record, without merging it with another stream. In the phone streams, all records have a reliable time point, and adding the time zone just adds some information. Perhaps adding a `timeOfDay` field with seconds since midnight could substitute for that? The new polling route + schema would be good as well.

yatharthranjan:
I agree. The Local date time+zone ID in UTC along with profile info should work from data analysis point of view. Timeofday as received from fitbit should also be included. And is the time and timeReceived both used here ?

blootsvoets:
Ok, the full list of fields for heart rate would then be `time`, `timeReceived`, `localTimeOfDay`, `timeInterval`, `heartRate` and there would be an additional timezone topic with a schema`time`, `offset` fields?

mbegalevibrent:
I was able to reach the fitbit team and was provided access to an undocumented but stable process for requesting UTC values. I propose we revert to the last proposed schemas discussed before identifying the UTC issue, work for you @blootsvoets @yatharthranjan?

yatharthranjan:
@mbegalevibrent that is great news. The current schemas in this PR are based on UTC timestamps. Please make any changes required. Also does this also include the UTC offset value ?

PR #119, opened 2018-05-03
Add Connector specs

### LINKED ISSUE #112: Separate connector definitions
In the RADAR-Schemas hierarchy, I have the feeling that `connector` should be a separate category from `passive`, `active`, `monitor` and `stream`. For example, we should not register the `connector` schemas during initialisation, because the Kafka Connector code will do that instead. If we try to do it, there will be an incompatibility between the schema that the connector provides and the one that we provide, because the Kafka Connector adds some connect-specific properties to the schema. Also, the connector does not have an app-provider, but it could have the name of the source connector.

Alternatively, we could add a flag in the specifications that says something like `register_schema: false`.

@yatharthranjan @nivemaham what do you think?

yatharthranjan:
- Add skeleton for connector source type
- Add schemas and specifications for fitbit connector
- ~~Add `output_topic` to `StreamDataTopic`~~
- Closes #112 
- @mbegalevibrent this PR will enable to register the topics for fitbit connector but the schemas will need to be registered from the connector. Please see #112

### DIFF
--- a/commons/connector/fitbit/fitbit_intraday_heart_rate.avsc
+++ b/commons/connector/fitbit/fitbit_intraday_heart_rate.avsc
@@ -1,13 +1,13 @@
 {
-  "namespace": "org.radarcns.passive.fitbit",
+  "namespace": "org.radarcns.connector.fitbit",
   "type": "record",
-  "name": "FitbitIntradaySteps",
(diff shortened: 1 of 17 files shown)


yatharthranjan:
I think we can merge for now since the kafka-connect approach will take some time. We can change the schema registration when its implemented.

PR #121, opened 2018-05-11
Add questionnaire completion log schema and specs

### LINKED ISSUE #120: Add Schema and specs for questionnaire completion log
The aRMT will send a questionnaire completion log that will help calculate the compliance

yatharthranjan:
closes #120

### DIFF
--- a/commons/active/questionnaire/questionnaire_completion_log.avsc
+++ b/commons/active/questionnaire/questionnaire_completion_log.avsc
@@ -6,6 +6,6 @@
   "fields": [
     { "name": "time", "type": "double", "doc": "Timestamp in UTC (s) when the questionnaire completion log is submitted." },
     { "name": "name", "type": "string", "doc": "Questionnaire name." },
-    { "name": "isComplete", "type": "string", "doc": "It reports the questionnaire completion status. Could be true or false." }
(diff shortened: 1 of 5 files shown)


commons/active/questionnaire/questionnaire_completion_log.avsc
```diff
@@ -0,0 +1,11 @@
+{
+  "namespace": "org.radarcns.active.questionnaire",
+  "type": "record",
+  "name": "QuestionnaireCompletionLog",
+  "doc": "Schema for reporting the completion status of a questionnaire. This will help in calculating the compliance.",
+  "fields": [
+    { "name": "time", "type": "double", "doc": "Timestamp in UTC (s) when the questionnaire completion log is submitted." },
+    { "name": "name", "type": "string", "doc": "Questionnaire name." },
+    { "name": "completionPercentage", "type": "double", "doc": "Percentage of the questionnaire completed. 0 for not at all complete and 100 for full completion."}
```
afolarin: I wonder if we should provide the option to include a NULL value say completion is not a value for a question type or we in the future include aRMT questions where there is no logical completion value (e.g. a hidden field)?

blootsvoets:
Perhaps the completion log can be moved to the `monitor` scope, because it's not actual user data but rather application metadata. Then the `timeCompleted` check can be added back in the validator.

PR #126, opened 2018-06-04
Add question id

yatharthranjan:


blootsvoets:
Would appreciate a bit longer wait for reviews.

yatharthranjan:
sorry, will keep in mind. Also we don't want it to be null in the future, is it possible somehow?

blootsvoets:
No unfortunately not. That’s why it’s good to plan ahead with these schemas. Or indeed accept that there are multiple incompatible versions of the schema, and force the update on existing installations.

PR #134, opened 2018-08-01
Updated gradle and dependencies

blootsvoets:


java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/registration/KafkaTopics.java
```diff
@@ -85,6 +74,15 @@ public boolean waitForBrokers(int brokers) throws InterruptedException, ZkExcept
                         activeBrokers, brokers, sleep);
             }
         }
+
+        String bootstrapServers = JavaConverters$.MODULE$
+                .seqAsJavaList(zkClient.getAllBrokersInCluster()).stream()
```
yatharthranjan: Is this for conversion between scala and java ? Maybe a comment here can be useful

java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/registration/KafkaTopics.java
```diff
@@ -60,10 +50,9 @@ public Object deserialize(byte[] bytes) throws ZkMarshallingError {
...
      * @throws InterruptedException when waiting for the brokers is interrepted.
-     * @throws ZkException if the Zookeeper instance cannot be reached or returns an unexpected
-     *                     result.
      */
-    public boolean waitForBrokers(int brokers) throws InterruptedException, ZkException {
+    public boolean waitForBrokers(int brokers) throws InterruptedException,
+            ZooKeeperClientException {
         boolean brokersAvailable = false;
         int sleep = 2;
         for (int tries = 0; tries < 10; tries++) {
```
yatharthranjan: Can we increase the timeout a bit? Because if we have a lot of data and we stop the cluster and then running the install script re-creates the brokers, then this times-out since kafka takes some time to load/recover all the data from snapshots.

PR #141, opened 2018-10-19
Add timeNotification and timeZone to questionnaire.avsc

mpgxvii:
Solves https://github.com/RADAR-base/RADAR-Questionnaire/issues/173


commons/active/questionnaire/questionnaire.avsc
```diff
@@ -6,6 +6,8 @@
   "fields": [
     { "name": "time", "type": "double", "doc": "Timestamp in UTC (s) when the questionnaire is submitted to the subject." },
     { "name": "timeCompleted", "type": "double", "doc": "Timestamp in UTC (s) when subject marks the questionnaire as." },
+    { "name": "timeNotification", "type": "double", "doc": "Timestamp in UTC (s) when the notification to complete the questionnaire is sent." },
+    { "name": "timeZone", "type": "string", "doc": "Local time zone that the device is running in." },
```
afolarin: @blootsvoets do you think this should be UTC offset? Rather than String `timeZone`

afolarin:
Also -- as essentially adding two new values (the same for all tasks I think) I don't think this is a backwards compatibility issue.

commons/active/questionnaire/questionnaire.avsc
blootsvoets: Yes, indeed, seconds from UTC. In pRMT the variable is called `offset` but that is a topic specifically about the time zone. Also, because these are new variables, we should set a default (`null`) for older values that do not have this variable. Perhaps here it could be:
```suggestion
    { "name": "timeNotification", "type": ["null", "double"], "doc": "Timestamp in UTC (s) when the notification to complete the questionnaire is sent.", "default": null },
    { "name": "timeZoneOffset", "type": ["null", "int"], "doc": "Local time zone offset from UTC (s).", "default": null },
```

blootsvoets: Yes, indeed, seconds from UTC. In pRMT the variable is called `offset` but that is a topic specifically about the time zone. Also, because these are new variables, we should set a default (`null`) for older values that do not have this variable. Perhaps here it could be:
```suggestion
    { "name": "timeNotification", "type": ["null", "double"], "doc": "Timestamp in UTC (s) when the notification to complete the questionnaire is sent.", "default": null },
    { "name": "timeZoneOffset", "type": ["null", "int"], "doc": "Local time zone offset from UTC (s).", "default": null },
```

blootsvoets: Yes, indeed, seconds from UTC. In pRMT the variable is called `offset` but that is a topic specifically about the time zone. Also, because these are new variables, we should set a default (`null`) for older values that do not have this variable. Perhaps here it could be:
```suggestion
    { "name": "timeNotification", "type": ["null", "double"], "doc": "Timestamp in UTC (s) when the notification to complete the questionnaire is sent.", "default": null },
    { "name": "timeZoneOffset", "type": ["null", "int"], "doc": "Local time zone offset from UTC (s).", "default": null },
```

blootsvoets: Yes, indeed, seconds from UTC. In pRMT the variable is called `offset` but that is a topic specifically about the time zone. Also, because these are new variables, we should set a default (`null`) for older values that do not have this variable. Perhaps here it could be:
```
    { "name": "timeNotification", "type": ["null", "double"], "doc": "Timestamp in UTC (s) when the notification to complete the questionnaire is sent.", "default": null },
    { "name": "timeZoneOffset", "type": ["null", "int"], "doc": "Local time zone offset from UTC (s).", "default": null },
```

blootsvoets: Yes, indeed, seconds from UTC. In pRMT the variable is called `offset` but that is a topic specifically about the time zone. Also, because these are new variables, we should set a default (`null`) for older values that do not have this variable. Perhaps here it could be:
```suggestion
    { "name": "timeNotification", "type": ["null", "double"], "doc": "Timestamp in UTC (s) when the notification to complete the questionnaire is sent.", "default": null },
    { "name": "timeZoneOffset", "type": ["null", "int"], "doc": "Local time zone offset from UTC (s).", "default": null },
```

blootsvoets: Yes, indeed, seconds from UTC. In pRMT the variable is called `offset` but that is a topic specifically about the time zone. Also, because these are new variables, we should set a default (`null`) for older values that do not have this variable. Perhaps here it could be:
```suggestion
    { "name": "timeNotification", "type": ["null", "double"], "doc": "Timestamp in UTC (s) when the notification to complete the questionnaire is sent.", "default": null },
    { "name": "timeZoneOffset", "type": ["null", "int"], "doc": "Local time zone offset from UTC (s).", "default": null },
```

blootsvoets:
We tried previously to add new values to existing topics, unfortunately that did lead to some backwards compatibility issues when an old schema ID was somehow invalidated by Kafka.  As long as the Schema Registry behaves though, it should not be a problem if a default is given.

blootsvoets:
We tried previously to add new values to existing topics, unfortunately that did lead to some backwards compatibility issues when an old schema ID was somehow invalidated by Kafka.  As long as the Schema Registry behaves though, it should not be a problem if a default is given.

Instead of adding variables, you could also consider adding a topic for the participants time zone (like https://github.com/RADAR-base/RADAR-Schemas/blob/master/commons/monitor/application/application_time_zone.avsc) and one for when notifications are delivered and activated. That way you have an independent measure of how notifications behave, also for notifications that are not acted upon.

blootsvoets:
We tried previously to add new values to existing topics, unfortunately that did lead to some backwards compatibility issues when an old schema ID was somehow invalidated by Kafka.  As long as the Schema Registry behaves though, it should not be a problem if a default is given.

Instead of adding variables, you could also consider adding a topic for the participants time zone (like https://github.com/RADAR-base/RADAR-Schemas/blob/master/commons/monitor/application/application_time_zone.avsc) and one for when notifications are delivered and activated. That way you have an independent measure of how notifications behave, also for notifications that are not acted upon.

commons/active/questionnaire/questionnaire.avsc
blootsvoets: Yes, indeed, seconds from UTC. In pRMT the variable is called `offset` but that is a topic specifically about the time zone. Also, because these are new variables, we should set a default (`null`) for older values that do not have this variable. Perhaps here it could be:
```suggestion
    { "name": "timeNotification", "type": ["null", "double"], "doc": "Timestamp in UTC (s) when the notification to complete the questionnaire is sent.", "default": null },
    { "name": "timeZoneOffset", "type": ["null", "int"], "doc": "Local time zone offset from UTC (s).", "default": null },
```

blootsvoets:
We tried previously to add new values to existing topics, unfortunately that did lead to some backwards compatibility issues when an old schema ID was somehow invalidated by Kafka.  As long as the Schema Registry behaves though, it should not be a problem if a default is given.

Instead of adding variables, you could also consider adding a topic for the participants time zone (like https://github.com/RADAR-base/RADAR-Schemas/blob/master/commons/monitor/application/application_time_zone.avsc) and one for when notifications are delivered and activated. That way you have an independent measure of how notifications behave, also for notifications that are not acted upon.

yatharthranjan:
I agree that we should have this in a different topic to avoid any problems in the backend. But if this info (Notification trigger time) is used as an identifier for each questionnaire then makes sense to add in the same schema. 

If we plan to use a different topic then the Notification time and delivery status info can also be sent by the FCM XMPP notification server to kafka. So the app really only needs to send the timezone offset then.

yatharthranjan:
I agree that we should have this in a different topic to avoid any problems in the backend. But if this info (Notification trigger time) is used as an identifier for each questionnaire then makes sense to add in the same schema. 

If we plan to use a different topic then the Notification time and delivery status info can also be sent by the FCM XMPP notification server to kafka. So the app really only needs to send the timezone offset then.

yatharthranjan:
I agree that we should have this in a different topic to avoid any problems in the backend. But if this info (Notification trigger time) is used as an identifier for each questionnaire then makes sense to add in the same schema. 

If we plan to use a different topic then the Notification time and delivery status info can also be sent by the FCM XMPP notification server to kafka. So the app really only needs to send the timezone offset then.

yatharthranjan:
I agree that we should have this in a different topic to avoid any problems in the backend. But if this info (Notification trigger time) is used as an identifier for each questionnaire then makes sense to add in the same schema. 

If we plan to use a different topic then the Notification time and delivery status info can also be sent by the FCM XMPP notification server to kafka. So the app really only needs to send the timezone offset then.

yatharthranjan:
I agree that we should have this in a different topic to avoid any problems in the backend. But if this info (Notification trigger time) is used as an identifier for each questionnaire then makes sense to add in the same schema. 

If we plan to use a different topic then the Notification time and delivery status info can also be sent by the FCM XMPP notification server to kafka. So the app really only needs to send the timezone offset then.

blootsvoets:
@yatharth for time zones, perhaps the schema from monitor/application_time_zone can be used, but with a different topic in the specification?

afolarin:
LGTM

afolarin:
@yatharthranjan @mpgxvii the suggested changes seem good, as is @blootsvoets suggested reuse of the timezone schema

The metadata question is a good one but, one question I had though was how the metadata-field should be structured.

yatharthranjan:
Yes the suggestion of reusing the timezone schema sounds good. 

@afolarin The metadata field can have the type `map` with `values` as `string`, so key value type information can be sent via it. @blootsvoets what do you think ?

blootsvoets:
Four downsides of a generic key-value map:
- any change in keys will yield a new CSV file in the output, i.e. harder to compare columns.
- the keys meanings are not documented in a schema
- less efficient storage and transmission, because the key names have to accompany the values. Binary encoding Avro does not store the key names.
- less efficient transmission of e.g. number or enum data, because it has to be transformed to string.
Upside:
- it’s more flexible

This is why the guidelines discourage such a generic approach. It’s also a reason that in Kafka, Avro is used over JSON in the first place. It is just a guideline, so it can be ignored in cases where there is a strong reason to not use it. Personally i’d prefer amending topics when there is a need or creating additional topics for alternative data schemas.

afolarin:
yes, no free lunch here. I guess we should weigh up the cost of future schema changes and the likelihood that such changes are needed or could be managed via a non-specified field. It may be prudent to hold filre on it as we haven't been inundated with schema update requests, but if it's very painful to update then perhaps we can think again.

yatharthranjan:
Okay seeing as many downsides, Maybe its better to add new topics for additional data. Thanks @blootsvoets for the insight. @afolarin recent experience in dealing with ESM's suggests that it is not exactly our typical questionnaire, maybe in future we can move ESM to a different schema than the other questionnaires so easy to update and does not break any others. What do you think ?

commons/active/questionnaire/questionnaire.avsc
```diff
@@ -6,6 +6,7 @@
   "fields": [
     { "name": "time", "type": "double", "doc": "Timestamp in UTC (s) when the questionnaire is submitted to the subject." },
     { "name": "timeCompleted", "type": "double", "doc": "Timestamp in UTC (s) when subject marks the questionnaire as." },
```
blootsvoets:
```suggestion
    { "name": "time", "type": "double", "doc": "Timestamp in UTC (s) when the questionnaire is started by the subject." },
    { "name": "timeCompleted", "type": "double", "doc": "Timestamp in UTC (s) when subject marks the questionnaire as completed." },
```
Doc update to avoid any ambiguities.

commons/active/questionnaire/questionnaire.avsc
```diff
@@ -6,6 +6,7 @@
   "fields": [
     { "name": "time", "type": "double", "doc": "Timestamp in UTC (s) when the questionnaire is submitted to the subject." },
```
blootsvoets:
```suggestion
     { "name": "time", "type": "double", "doc": "Timestamp in UTC (s) when the questionnaire is started by the subject." },
```

PR #143, opened 2018-10-25
Added PPG schemas

blootsvoets:
Schemas for a new plugin (https://github.com/rate-af/radar-android-ppg/tree/radar-plugin) to measure PPG using a phone camera. This is actively going to be used in BigData@Heart/RATE-AF but seems useful as a general RADAR plugin.

### DIFF
--- a/specifications/passive/rateaf_ppg-1.0.0.yml
+++ b/specifications/passive/rateaf_ppg-1.0.0.yml
@@ -1,0 +1,17 @@
+#====================================== Android Phone Sensors =====================================#
+vendor: RATE-AF
+model: PPG
+version: 0.1.0
+data:
+  #Phone sensors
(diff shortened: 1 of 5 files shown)


specifications/passive/rateaf_ppg-1.0.0.yml
```diff
@@ -0,0 +1,17 @@
+#====================================== Android Phone Sensors =====================================#
+vendor: RATE-AF
+model: PPG
+version: 0.1.0
```
yatharthranjan: 1.0.0 ?

blootsvoets:
The Samsung Sp02 measurement indeed internally probably uses a similar measure. I suggest that if we write a plugin for that, it will be separate from this one.

PR #144, opened 2018-10-29
Print created topics

blootsvoets:
Print created and existing topics during initialisation.


java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/registration/KafkaTopics.java
```diff
@@ -127,8 +127,18 @@ public boolean createTopics(Stream<String> topics, int partitions, short replica
         try {
             Set<String> existingTopics = kafkaClient.listTopics().names().get();
 
+            logger.info("Creating topics. Topics marked with [*] already exist.");
```
nivemaham: It only logs about the topics that already exist. Maybe we should add a line to print the other value too?

blootsvoets: It list all topics, but topics that already exist will be prepended with `[*]` and topics that do not yet exist will be prepended with `[ ]`. Do you mean it should also list topics that are not requested in the current "create topics" query? That would need some more adjustments, and possibly a new pull request.

blootsvoets: Also, it can be achieved with just a `docker-compose exec rest-proxy-1 curl  localhost:9092/topics`

PR #150, opened 2018-11-13
WIP: Add Fitbit activity log records

blootsvoets:


commons/connector/fitbit/fitbit_activity_heart_rate.avsc
```diff
@@ -0,0 +1,18 @@
+{
+  "namespace": "org.radarcns.connector.fitbit",
+  "name": "FitbitActivityHeartRate",
+  "type": "record",
+  "doc": "Aggregate heart rate as measured over an activity. The heart rates are divided amongst so-called heart rate zones. A heart rate zone is a range of heart rates and it does not overlap with other heart rate zones. These zones are called, in order of increasing heart rate, 'Out of Range', 'Fat Burn', 'Cardio', and 'Peak'.",
```
afolarin: do we want to qualify these zones explicitly https://blog.fitbit.com/heart-rate-zones/? Even if they may change in the future the zones or bins.

afolarin: on second thoughts perhaps better left in Fitbit docs.

blootsvoets: Ah I already updated the docs, is the following ok?
> Aggregate heart rate as measured over an activity. The heart rates are divided amongst so-called heart rate zones. A heart rate zone is a range of heart rates and it does not overlap with other heart rate zones. These zones are called, in order of increasing heart rate, 'Out of Range', 'Fat Burn', 'Cardio', and 'Peak'. Fitbit uses the common formula of 220 minus your age to estimate your maximum heart rate. The 'Out of Range' zone has heart rates up to 50 percent of maximum heart rate, the 'Fat Burn' has heart rates up to 70 percent of maximum heart rate, 'Cardio' up to 85 percent of maximum heart rate and the 'Peak' range has heart rates higher than that.

blootsvoets:
Used in RADAR-base/RADAR-REST-Connector#13.

commons/connector/fitbit/fitbit_activity_log_record.avsc
```diff
@@ -0,0 +1,26 @@
...
+    {"name": "timeZoneOffset", "type": "int", "doc": "Time zone offset compared to UTC (s)."},
+    {"name": "timeLastModified", "type": "double", "doc": "Time when the record was last modified, time since the Unix Epoch (s)."},
+    {"name": "duration", "type": "float", "doc": "Duration of the activity (s)."},
+    {"name": "durationActive", "type": "float", "doc": "Duration of the part of the activity where the participant was active (s)."},
+    {"name": "id", "type": "long", "doc": "Activity log ID."},
+    {"name": "name", "type": ["null", "string"], "doc": "Activity name. Null if none or unknown.", "default": null},
+    {"name": "logType", "type": ["null", "string"], "doc": "Type of log, e.g. manual. Null if unknown.", "default": null},
+    {"name": "type", "type": "long", "doc": "Fitbit-generated ID of the type of activity."},
+    {"name": "source", "type": ["null", "FitbitSource"], "doc": "Source of the data. Null if unknown.", "default": null},
+    {"name": "manualDataEntry", "type": ["null", "FitbitManualDataEntry"], "doc": "Indicates what values of this record are entered manually. Null if unknown.", "default": null},
```
afolarin: is there some ambiguity between manualDataEntry=null, i.e. if it is supplied from the tracker ?

PR #160, opened 2019-01-24
Empatica spec

sboettcher:
Added the sensor_status topic to the empatica specification, without it the topic is not created by the radar-schemas-tools and the pRMT app won't send the data (throws unblocking exception).

I put it into a new patch version, wasn't sure if that is OK or if it should just be added to v1.0.0.


yatharthranjan:
Hi Sebastian,
For a new topic the version can be 1.X.0, so in this case 1.1.0. Otherwise looks good to me.

Thanks

PR #161, opened 2019-01-25
Add timeNotification to completion log

mpgxvii:


yatharthranjan:
Can we add attemptNumber too? If it is a planned feature?

mpgxvii:
Does that mean we have to store it as a property of each task?

PR #162, opened 2019-02-07
Validate that a conversion to Kafka will succeed

blootsvoets:
Related to RADAR-base/RADAR-HDFS-Sink-Connector#19, validate that all records in the repository can be converted to Confluent/Kafka Schema and back again.


java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/validation/rules/RadarSchemaRules.java
```diff
@@ -201,6 +202,27 @@ public SchemaFieldRules getFieldRules() {
                 messageSchema("\"" + TIME_RECEIVED + "\" is allow only in PASSIVE schemas."));
     }
 
+    @Override
+    public Validator<Schema> validateAvroData() {
+        return schema -> {
+            AvroDataConfig avroConfig = new AvroDataConfig.Builder()
+                    .with("connect.meta.data", false)
+                    .with("schemas.cache.config", 10)
+                    .with("enhanced.avro.schema.support", true)
```
yatharthranjan: We can probably use constants from the AvroDataConfig class like `AvroDataConfig.CONNECT_META_DATA_CONFIG`.

PR #163, opened 2019-02-27
Update to Java 11, schema topic management, PMD update

blootsvoets:
Big update to facilitate further upgrading. The tools now depend on Java 11. This was needed for  some small features, but will keep images in line with future Java development.

Main new feature: it is now possible to back up the entire _schemas topic and to replace it with a correct one if it has become corrupted. This change was made to avoid corruptions as we encounter in RADAR-base once in a while. It is managed by 

This should be accompanied by an update to RADAR-Docker, where before the schema registry is started and schema registration is done, a `_schemas` topic check is done:
```
radar-schemas-tools schema-topic --ensure --brokers 3 --replication 3 -f schema-topic.json zookeeper-1:2181,zookeeper-2:2181,zookeeper-3:2181
```
Alternatively, there could be a script to bring all Avro-related services down and run this check. A regular backup using a timer should be made once the stack is running.


java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/registration/SchemaTopicManager.java
```diff
@@ -0,0 +1,506 @@
...
+     * Read a backup from the _schemas topic and store it. This backup only includes the actual
+     * schemas, not configuration changes or NOOP messages.
+     * @param timeout time to wait for first schema records to become available.
+     * @throws SerializationException if a message in the topic cannot be read
+     * @throws ExecutionException if the topic configuration cannot be read
+     * @throws InterruptedException if the process was interrupted before finishing
+     * @throws RuntimeException storage failure or any other error.
+     * @throws IllegalStateException if this manager was not initialized
+     */
+    public void makeBackup(Duration timeout)
```
blootsvoets: FYI: entry point to make a backup

java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/registration/SchemaTopicManager.java
```diff
@@ -0,0 +1,506 @@
...
+     * @param timeout time to wait for first schema records to become available.
+     * @throws SerializationException if a message in the topic cannot be read or written
+     * @throws ExecutionException if the topic configuration cannot be read or written
+     * @throws InterruptedException if the process was interrupted before finishing
+     * @throws RuntimeException storage failure or any other error.
+     * @throws IllegalStateException if this manager was not initialized or schema registry was
+     *                               running
+     * @throws KeeperException if no connection with Zookeeper could be made.
+     */
+    public void ensure(short replication, Duration timeout)
```
blootsvoets: FYI: entrypoint to check `_schemas` topic validity and replace it with a backup if needed.

java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/registration/SchemaTopicManager.java
```diff
@@ -0,0 +1,506 @@
...
+
+    /**
+     * Restores the _schemas from backup.
+     * @throws RuntimeException storage failure or any other error.
+     * @throws ExecutionException if the topic configuration cannot be written
+     * @throws IllegalStateException if this manager was not initialized or if the
+     *                               schema registry is running.
+     * @throws KeeperException if no connection with Zookeeper could be made.
+     */
+    public void restoreBackup(short replication)
```
blootsvoets: FYI: entry point to completely restore a backup. This can only be called if the `_schemas` topic was deleted.

java-sdk/radar-schemas-tools/src/main/java/org/radarcns/schema/registration/SchemaTopicBackup.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    /**
+     * Put settings.
+     *
+     * @param settings Kafka topic config settings
+     */
+    public void putAll(@NotNull Map<? extends String, ? extends String> settings) {
+        this.settings.putAll(settings);
+        if (!Objects.equals(settings.get(CLEANUP_POLICY_CONFIG), CLEANUP_POLICY_COMPACT)) {
```
nivemaham: why do we force set this to `CLEANUP_POLICY_COMPACT`?

blootsvoets: Because if the cleanup policy is `DELETE`, data will be deleted after a set amount of time. With compact strategy, Kafka should not actively delete any data in the topic if it has a unique key.

PR #164, opened 2019-03-13
Prevent application status int overflow

blootsvoets:
The number of records cached is a bit theoretical (but not entirely out of reach), because it would require about 200 GB of storage. However, at 1000 Hz, recordsSent will exceed integer maximum value after 24 days, which is quite possible.

### DIFF
--- a/commons/monitor/application/application_record_counts.avsc
+++ b/commons/monitor/application/application_record_counts.avsc
@@ -5,8 +5,8 @@
   "doc": "Number of records cached or created.",
   "fields": [
     { "name": "time", "type": "double", "doc": "Device timestamp in UTC (s)." },
-    { "name": "recordsCached", "type": ["null", "int"], "doc": "Number of records currently being cached.", "default": null },
-    { "name": "recordsSent", "type": "int", "doc": "Number of records sent since application start." },
+    { "name": "recordsCached", "type": ["null", "long"], "doc": "Number of records currently being cached.", "default": null },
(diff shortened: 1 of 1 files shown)


commons/monitor/application/application_record_counts.avsc
```diff
@@ -5,8 +5,8 @@
   "doc": "Number of records cached or created.",
   "fields": [
     { "name": "time", "type": "double", "doc": "Device timestamp in UTC (s)." },
-    { "name": "recordsCached", "type": ["null", "int"], "doc": "Number of records currently being cached.", "default": null },
-    { "name": "recordsSent", "type": "int", "doc": "Number of records sent since application start." },
+    { "name": "recordsCached", "type": ["null", "long"], "doc": "Number of records currently being cached.", "default": null },
+    { "name": "recordsSent", "type": "long", "doc": "Number of records sent since application start." },
```
yatharthranjan: should we add long instead of replacing int ?

blootsvoets: IIRC, there was an issue earlier with AvroConnect if more than two union members were used. I don't recall the specifics though...

yatharthranjan: Will the recordsCached be backward compatible ?

blootsvoets: In JSON encoding: yes, ints will be correctly be parsed as longs.
In binary encoding: yes, ints will be correctly parsed as longs, because of their encoding format (variable length zig-zag encoding).

PR #165, opened 2019-03-19
Update thinc_it_spotter.avsc

jorgecid:
add mean latency to quarter summary


commons/active/thincit/thinc_it_spotter.avsc
```diff
@@ -17,6 +17,7 @@
       "name": "ThincItSpotterQuarterSummary",
       "doc": "Trial summary for SymbolCheck assessment.",
       "fields": [
+        { "name": "meanLatency", "type": "int", "doc": "Mean latency of correct responses expressed in milliseconds." },
```
yatharthranjan: Please add null as a default type.

```suggestion
        { "name": "meanLatency", "type": ["null", "int"], "doc": "Mean latency of correct responses expressed in milliseconds.", "default": null },
```

PR #187, opened 2019-10-23
Updated altoida topics and added summary

blootsvoets:
- Updated Altodia schema names to match other sensors
  - magnetometer -> magnetic field
  - touchscreen -> tap
- Added a AltoidaSummary schema with subschemas to map Altoida's `CSV-export.csv`.

Note: pending documentation on the `nmi`, `groundTruth`, `result`, `bitMetrics` and `dotMetrics` fields.


commons/connector/upload/altoida/altoida_tapping_test_aggregate.avsc
```diff
@@ -0,0 +1,15 @@
+{
+  "namespace": "org.radarcns.connector.upload.altoida",
```
nivemaham: I know that we skipped integrating the CSV_Export file. Are these part of already sent zip file?

nivemaham:
I think we can also use this opportunity to add the additional files sent by Altoida. The EYE, BLINK and TOUCH ?

PR #193, opened 2019-11-25
added epi questionnaire to aRMT specifications -> aRMT-1.4.4

sboettcher:
questionnaire definition URLs dependent on https://github.com/RADAR-base/RADAR-REDCap-aRMT-Definitions/pull/77

### DIFF
--- a/specifications/active/aRMT-1.4.4.yml
+++ b/specifications/active/aRMT-1.4.4.yml
@@ -1,0 +1,83 @@
+name: aRMT
+vendor: RADAR
+model: aRMT-App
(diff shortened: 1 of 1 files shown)


sboettcher:
Hm ok, so in this case it would be 1.5.0, since there are new topics?

yatharthranjan:
@sboettcher yes exactly! 💯

sboettcher:
OK, in that case I think it will be better to wait with merging, until we have added all topics that we need. I am not sure currently if there are any more that we need to add...

PR #194, opened 2019-11-29
move bit and dot metrics and domainresult to separate topics

nivemaham:
move bit and dot metrics and domainresult to separate topics


nivemaham:
Good question. maybe we should make a habit of it for the future. So far we haven't used the version on uploader. So it doesn't affect so far.

PR #211, opened 2020-02-06
Add physilog data schema

nivemaham:
add schemas for physilog binary data reference


commons/connector/upload/physilog/physilog_binary_data.avsc
```diff
@@ -0,0 +1,12 @@
+{
+  "namespace": "org.radarcns.connector.upload.physilog",
+  "type": "record",
+  "name": "PhysilogBinaryData",
```
yatharthranjan: Can this be any data in binary format? Or will this always be image data?

PR #212, opened 2020-02-10
Release 0.5.7

nivemaham:
**Changes from Release 0.5.6**
- Adds schema and spec for physilog 5


commons/connector/upload/physilog/physilog_binary_data.avsc
```diff
@@ -0,0 +1,12 @@
+{
+  "namespace": "org.radarcns.connector.upload.physilog",
+  "type": "record",
+  "name": "PhysilogBinaryData",
```
yatharthranjan: Don't know much about this but would `PhysilogBinaryDataReference` be better suited name? And same for the topic in specs?

PR #215, opened 2020-03-16
Add RAPID questionnaires to active schema specs

mpgxvii:
- Adds QIDS, WSAS, Treatment Status, Sleep Questions, and Social Activities questionnaires for RAPID study to active schema specs


specifications/active/aRMT-1.5.2.yml
```diff
@@ -0,0 +1,138 @@
+name: aRMT
+vendor: RADAR
+model: aRMT-App
+version: 1.5.2
```
yatharthranjan: Can we version it as 1.6.0. Since we are adding topics this will be considered a minor version update and not a patch version. 
Thanks

specifications/active/aRMT-1.5.2.yml
```diff
@@ -0,0 +1,138 @@
...
+  - type: WSAS
+    topic: questionnaire_wsas
+    value_schema: .active.questionnaire.Questionnaire
+    questionnaire_definition_url: https://raw.githubusercontent.com/RADAR-base/RADAR-REDCap-aRMT-Definitions/master/questionnaires/wsas/wsas_armt.json
+  - type: TREATMENT_STATUS
+    topic: questionnaire_treatment_status
+    value_schema: .active.questionnaire.Questionnaire
+    questionnaire_definition_url: https://raw.githubusercontent.com/RADAR-base/RADAR-REDCap-aRMT-Definitions/master/questionnaires/treatment_status/treatment_status_armt.json
+  - type: SLEEP_QUESTIONS
+    topic: questionnaire_sleep_questions
```
yatharthranjan: What are these? Can we name these a bit better since sleep questions is very generic.

specifications/active/aRMT-1.5.2.yml
```diff
@@ -0,0 +1,138 @@
...
+    value_schema: .active.questionnaire.Questionnaire
+    questionnaire_definition_url: https://raw.githubusercontent.com/RADAR-base/RADAR-REDCap-aRMT-Definitions/master/questionnaires/sleep_questions/sleep_questions_armt.json
+  - type: SOCIAL_ACTIVITIES
+    topic: questionnaire_social_activities
```
yatharthranjan: Same here. Maybe add the study abbreviation in the name.

specifications/active/aRMT-1.5.2.yml
mpgxvii: I'm not sure. These were directly taken from REDCap. Maybe we can add `RAPID`? Or we can ask their team to rename it?

yatharthranjan: As these are just topic names, I think we can decide ourselves. Yes, I think something like `questionnaire_rapid_sleep_questions` should be fine.

yatharthranjan (approved):
LGTM

PR #237, opened 2020-08-03
Update aRMT specs

mpgxvii:
- Restore deleted `aRMT-1.6.1.yml`
- Add `CNS_COVID19_FOLLOWUP_V3` (CNS), `ART_COVID19_FOLLOWUP` (ART), `MEDICATION_USE` (ART) questionnaires to schema specs

PR #238, opened 2020-08-07
Release 0.5.12

mpgxvii:
Changes since version 0.5.11.1:

- Add new questionnaires to aRMT specifications (#237)
- Streamline docker files to be more similar
- Fix help message

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:11-jdk-slim
+FROM openjdk:11-jdk-slim as builder
 
 RUN mkdir -p /code/java-sdk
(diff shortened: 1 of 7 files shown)


specifications/active/aRMT-1.6.1.yml
```diff
@@ -0,0 +1,138 @@
+name: aRMT
```
nivemaham, line 1: any reason why these two versions of the spec can't be merged into one version?

mpgxvii: Because this was accidentally deleted, but it was released in a previous version.

PR #247, opened 2020-10-30
Schemas for Garmin Health API support

yatharthranjan:
I have added the schemas which are mapped 1:1 to what is provided by Garmin's health API.

~~The consideration right now is the use of `array` and `map` types (fields starting with `timeOffset*`).
These look fine in nested data formats (like json and avro), but i think will not be great for flat formats such as CSV. An alternative is to create an unnested schema and just produce each entry of the map as a separate record in kafka.~~

Let me know what you guys think @afolarin , @blootsvoets.

EDIT- Ok, i have now moved all the maps and arrays to individual records to avoid nesting.


commons/push/integration/garmin/garmin_push_activity_details.avsc
```diff
@@ -0,0 +1,46 @@
+{
+  "namespace": "org.radarcns.push.integration.garmin",
+  "name": "GarminActivityDetails",
+  "type": "record",
+  "doc": "Fitness activity details summaries represent detailed information about discrete fitness activities, such as running or swimming, that are specifically and intentionally started by the user on their device. All wellness data, like steps and distance, contained in the activity are already represented in the Daily summary and in the corresponding Epoch summaries, so Activity Detail summaries should only be used for programs that wish to treat specific activity types in different ways, such as giving the user extra credit for going swimming three times in the same week. Activity details summaries include all data recorded by the device as part of the Fitness Activity, including GPS coordinates and all recorded sensor data. ",
+  "fields": [
+    {"name": "summaryId", "type": "string", "doc": "Unique identifier for the summary."},
+    {"name": "summary", "type": {
+      "name": "GarminActivityDetailsSummary",
```
blootsvoets, line 9: Would you consider moving all these fields one level up, to avoid nesting?

commons/push/integration/garmin/garmin_push_activity_details.avsc
```diff
@@ -0,0 +1,46 @@
...
+        {"name": "manual", "type": ["null", "boolean"], "doc": "Indicates that the activity was manually entered directly on the Connect site. This property will only exist for manual activities.", "default": null},
+        {"name": "deviceName", "type": ["null", "string"], "doc": "Only Fitness Activities are associated with a specific Garmin device rather than the user’s overall account. If a user wears two devices at once during the same time and starts a Fitness Activity on each then both will generate separate Activity summaries with two different deviceNames. Always ‘unknown’ for manually created activities.", "default": null}
+      ]
+      }
+    },
+    {"name": "samples", "type": { "type": "array", "items" : "GarminActivityDetailsSample", "default": [] }, "doc": "Each activity detail contains an optional list of samples. The samples list will be empty if the activity is manual or details are not supported by the device. Samples may be as frequent as once per second, and values should be considered valid until the next sample."}
```
blootsvoets, line 44: Each of the samples should get their own record. You can add the summaryId to the samples to have a reference

blootsvoets (commented):
I just briefly looked at this, will look at it further next week

commons/push/integration/garmin/garmin_push_activity_details.avsc
```diff
@@ -0,0 +1,37 @@
...
+  "fields": [
+    {"name": "summaryId", "type": "string", "doc": "Unique identifier for the summary."},
+    {"name": "time", "type": "double", "doc": "Start time of the activity in seconds since January 1, 1970, 00:00:00 UTC (Unix timestamp)."},
+    {"name": "timeReceived", "type": "double", "doc": "Time that this record was collected by a service in seconds since the Unix Epoch (s)."},
+    {"name": "startTimeOffsetInSeconds", "type": ["null", "int"], "doc": "Offset in seconds to add to time to derive the 'local' time of the device that captured the data.", "default": null},
+    {"name": "activityType", "type": ["null", "string"], "doc": "Text description of the activity type. See Appendix A for a complete list.", "default": null},
```
yatharthranjan, line 11: There is a very long list of valid activity types. Not sure where is the best place to put it.

blootsvoets: In that case a URL to the activity types could work? You could give two or three examples inline, like e.g. walking, running?

commons/push/integration/garmin/garmin_push_activity_details.avsc
```diff
@@ -5,42 +5,33 @@
...
-      ]
-      }
-    },
-    {"name": "samples", "type": { "type": "array", "items" : "GarminActivityDetailsSample", "default": [] }, "doc": "Each activity detail contains an optional list of samples. The samples list will be empty if the activity is manual or details are not supported by the device. Samples may be as frequent as once per second, and values should be considered valid until the next sample."}
-  ]
+    {"name": "time", "type": "double", "doc": "Start time of the activity in seconds since January 1, 1970, 00:00:00 UTC (Unix timestamp)."},
+    {"name": "timeReceived", "type": "double", "doc": "Time that this record was collected by a service in seconds since the Unix Epoch (s)."},
+    {"name": "startTimeOffsetInSeconds", "type": ["null", "int"], "doc": "Offset in seconds to add to time to derive the 'local' time of the device that captured the data.", "default": null},
+    {"name": "activityType", "type": ["null", "string"], "doc": "Text description of the activity type. See Appendix A for a complete list.", "default": null},
+    {"name": "durationInSeconds", "type": ["null", "int"], "doc": "Length of the monitoring period in seconds.", "default": null},
```
blootsvoets, line 12: For these fields and the next, I’d opt for moving the unit out of the field name and just have it in the doc, like in the schema’s of other integrations.

commons/push/integration/garmin/garmin_push_activity_details.avsc
```diff
@@ -0,0 +1,37 @@
+{
+  "namespace": "org.radarcns.push.integration.garmin",
```
blootsvoets, line 2: That this is an integration comes automatically. Would rename to `org.radarcns.push.garmin`

commons/push/integration/garmin/garmin_push_epoch_summary.avsc
```diff
@@ -0,0 +1,22 @@
...
+    {"name": "activeKilocalories", "type": ["null", "int"], "doc": "Active kilocalories (dietary calories) burned during the monitoring period. This includes only the calories burned by the activity and not calories burned as part of the basal metabolic rate (BMR).", "default": null},
+    {"name": "met", "type": ["null", "double"], "doc": "MET (Metabolic Equivalent of Task) value for the active time for this activity type. Metabolic Equivalent of Task (MET) is an official measure of activity intensity. Garmin’s calculation of MET is an estimation based on the biometric data provided (height, weight, date of birth, gender) and improves in accuracy if heart rate data is also captured. The following linked document hosted by the US Centers for Disease Control and Prevention provides detailed information on MET and physical activity intensity: http://www.cdc.gov/nccdphp/dnpa/physical/pdf/PA_Intensity_table_2_1.pdf", "default": null},
```
blootsvoets, line 17: Would prefer that this acronym is written out `metabolicEquivalentOfTask`.

commons/push/integration/garmin/garmin_push_heart_rate_sample.avsc
```diff
@@ -0,0 +1,12 @@
+{
+  "namespace": "org.radarcns.push.integration.garmin",
+  "name": "GarminHeartRateSample",
+  "type": "record",
+  "doc": "Mapping between time (in seconds) to a heart rate value recorded for the time, in beats per minute. Each entry is a representative sample of the previous 15 seconds from the given time. Lack of entry for a given time should be interpreted as no data available. These are accompanied with the Dailies data and can be associated using the summaryId.",
+  "fields": [
+    {"name": "summaryId", "type": "string", "doc": "Unique identifier for the summary that this sample is associated with."},
+    {"name": "time", "type": "double", "doc": "Start time of the sample in seconds since January 1, 1970, 00:00:00 UTC (Unix timestamp)."},
+    {"name": "timeReceived", "type": "double", "doc": "Time this sample was recieved by the push service in seconds since January 1, 1970, 00:00:00 UTC (Unix timestamp)."},
+    {"name": "heartrate", "type": ["null", "double"], "doc": "Heart rate in beats per minute.", "default": null}
```
blootsvoets, line 10:
```suggestion
    {"name": "heartRate", "type": ["null", "double"], "doc": "Heart rate in beats per minute.", "default": null}
```

commons/push/integration/garmin/garmin_push_stress_level_sample.avsc
```diff
@@ -0,0 +1,12 @@
...
+  "fields": [
+    {"name": "summaryId", "type": "string", "doc": "Unique identifier for the summary that this sample is associated with."},
+    {"name": "time", "type": "double", "doc": "Start time of the sample in seconds since January 1, 1970, 00:00:00 UTC (Unix timestamp)."},
+    {"name": "timeReceived", "type": "double", "doc": "Time this sample was recieved by the push service in seconds since January 1, 1970, 00:00:00 UTC (Unix timestamp)."},
+    {"name": "stressLevel", "type": ["null", "double"], "doc": "Stress level value recorded for the time..", "default": null}
```
blootsvoets, line 10: Is this value derived in any documented way? Could be a description or a URL?

commons/push/integration/garmin/garmin_push_user_metrics.avsc
```diff
@@ -0,0 +1,14 @@
...
+  "namespace": "org.radarcns.push.integration.garmin",
+  "name": "GarminUserMetrics",
+  "type": "record",
+  "doc": "User Metrics are per-user calculations performed by Garmin based on the underlying data uploaded from the user’s device. This data can be specific to a single device and field availability is dependent on device model support. Unlike other summaries, User Metrics are associated only with a calendar date, not a specific time frame, and only the most recent value for any fields is presented to the user. Each metric directly corresponds to the similarly named field found in Garmin Connect.",
+  "fields": [
+    {"name": "summaryId", "type": "string", "doc": "Unique identifier for the summary."},
+    {"name": "time", "type": "double", "doc": "Start time of the activity in seconds since January 1, 1970, 00:00:00 UTC (Unix timestamp)."},
+    {"name": "timeReceived", "type": "double", "doc": "Time that this record was collected by a service in seconds since the Unix Epoch (s)."},
+    {"name": "calendarDate", "type": ["null", "string"], "doc": "The calendar date this summary would be displayed on in Garmin Connect. The date format is ‘yyyy-mm-dd’.", "default": null},
```
blootsvoets, line 10: Just `date` would also be fine. Please clarify in doc that this is the date in the local time zone of the garmin connect.

commons/push/integration/garmin/garmin_push_activity_details.avsc
```diff
@@ -0,0 +1,37 @@
...
+    {"name": "maxRunCadenceInStepsPerMinute", "type": ["null", "double"], "doc": "Maximum Run Cadence in Steps per minute.", "default": null},
+    {"name": "activeKilocalories", "type": ["null", "int"], "doc": "Active kilocalories (dietary calories) burned through actual movement and activity during the monitoring period.", "default": null},
+    {"name": "averageSpeedInMetersPerSecond", "type": ["null", "double"], "doc": "Average Speed in meters per second.", "default": null},
+    {"name": "maxSpeedInMetersPerSecond", "type": ["null", "double"], "doc": "Maximum Speed in meters per second.", "default": null},
+    {"name": "averageSwimCadenceInStrokesPerMinute", "type": ["null", "double"], "doc": "Average Swim Cadence in strokes per minute.", "default": null},
+    {"name": "averagePaceInMinutesPerKilometer", "type": ["null", "double"], "doc": "Average Pace in minutes per kilometer.", "default": null},
+    {"name": "maxPaceInMinutesPerKilometer", "type": ["null", "double"], "doc": "Maximum Pace in minutes per kilometer.", "default": null},
```
blootsvoets, line 23: All of these values and the following do not have enough digits to warrant double precision, float should be sufficient: [it can store 6 digits exactly](https://www.exploringbinary.com/decimal-precision-of-binary-floating-point-numbers/). Please make this evaluation for other records as well.

blootsvoets (changes requested):
Please find some comments inline. General comment for all records: please move the `summaryId` to after `time` and `timeReceived`, to match other records.

commons/push/integration/garmin/garmin_push_activity_details.avsc
yatharthranjan: No, there is no URL. It is in the documentation provided as a PDF file. I copy and put the list as a comma-separated file in this repo and link it here?

blootsvoets: If it is a public PDF, we could publish it as part of this repository and link to it that way?

yatharthranjan: Unfortunately, it is not public, it is a confidential document provided only through the dev portal available after you have access to the key

blootsvoets: Hmm, in that case, perhaps the entire list does need to be listed here... Alternatively, you use an enum to explicitly state what activity it is, but then if new activity types are added, the list needs to be updated to take that into account.

commons/push/integration/garmin/garmin_push_stress_level_sample.avsc
yatharthranjan: The only documentation is as mentioned in the record-level doc - `These numbers are derived based on a combination of many device sensors and will automatically adjust to the wearer of the device and gain accuracy over time as the stress algorithms learn the user’s natural biometric norms.`

yatharthranjan:
Ok, I have made the suggested changes, please take look. Thanks

blootsvoets (approved):
LGTM, thanks for the changes.

PR #248, opened 2020-11-02
Add specifications for Garmin Health API

yatharthranjan:
not sure if anything else is also required since this is adding a new folder under the specifications.


blootsvoets (commented):
The spec looks fine, but the radar-schemas-tools should also be adapted to scan push repositories and create the specified topics. Perhaps also add the version number to the file name.

blootsvoets (approved):
LGTM. Just to check, have you verified that this causes kafka-init to register the correct sources?

yatharthranjan:
@blootsvoets i checked creation of topics and schemas. I did not check the adding of source types in management portal.

PR #249, opened 2020-11-06
Update Altoida summary metrics schema

mpgxvii:
- Make `audioLowReactions` nullable since this is only present in the `DOT` test and not in the `BIT` test
- Remove `groundTruth` since this is not present anymore


nivemaham (approved):

mpgxvii:
> shouldn't the null be quoted?
I looked at the other schemas, they don't seem to have quotes. The test is now passing locally, just waiting for travis to pass as well.

PR #251, opened 2021-01-11
Update schemas specs for RAPID study

mpgxvii:
- Add back deleted RAPID questionnaire schemas specs caused by a merging issue (schema spec `0.9.1`)
- Also syncs `dev` with `master`


mpgxvii:
According to the clinical team, new changes are not needed.

PR #252, opened 2021-02-22
Update questionnaire_application_interaction_event schema

mpgxvii:
- Add `metadata` field to questionnaire_application_interaction_event schema
- Add `QUESTIONNAIRE_TRIGGERED` event type


commons/monitor/questionnaire/questionnaire_application_interaction_event.avsc
```diff
@@ -7,13 +7,14 @@
...
       "type": "enum",
-      "symbols": ["NOTIFICATION_OPEN", "APP_OPEN", "QUESTIONNAIRE_STARTED", "QUESTIONNAIRE_FINISHED", "QUESTIONNAIRE_CANCELLED", "OTHER", "UNKNOWN", "RECORDING_STARTED", "RECORDING_STOPPED"]
+      "symbols": ["NOTIFICATION_OPEN", "APP_OPEN", "QUESTIONNAIRE_STARTED", "QUESTIONNAIRE_FINISHED", "QUESTIONNAIRE_CANCELLED", "OTHER", "UNKNOWN", "RECORDING_STARTED", "RECORDING_STOPPED", "QUESTIONNAIRE_TRIGGERED"]
       },
       "doc": "Questionnaire app activity usage event type.",
       "default": "UNKNOWN"
     },
-    { "name": "questionnaireName", "type": ["null", "string"], "doc": "Name of the questionnaire.", "default": null }
+    { "name": "questionnaireName", "type": ["null", "string"], "doc": "Name of the questionnaire.", "default": null },
+    { "name": "metadata", "type": ["null", "string"], "doc": "Event metadata.", "default": null }
```
yatharthranjan, line 18: I would also add to the doc that, for the `QUESTIONNAIRE_TRIGGERED ` event, this will be used to propagate any metadata received in the data message payload.

blootsvoets: Indeed. Also in doc, what format this metadata will have.

yatharthranjan: @blootsvoets it can be a Map with any K,V pairs based on the output of the inference. This is just so we can know which questionnaire was triggered by which model and params. But i agree @mpgxvii, would be good to include this in the docs as well.

blootsvoets: If this list of params is more or less fixed, you could opt to just make this an actual K,V map (`{"type": "map","values":["null","string"]}`. The output CSV files will then put data with and without these parameters into separate files. Alternatively, please specify the syntax of the K,V map (e.g., JSON).

yatharthranjan: It will not be fixed and could vary with the model used for analysis. I agree we can perhaps make this a map.

mpgxvii: Okay sure I'll make the changes

yatharthranjan:
Can you merge the latest dev to fix the dependency issue for the CI to pass successfully?

blootsvoets (approved):
LGTM

PR #253, opened 2021-03-16
Publish to maven central

blootsvoets:
- Publish radar-schemas to Maven Central under `org.radarbase`
- Move all classes to `org.radarbase`, except the schema's.
- Remove bintray and artifactory


yatharthranjan (approved):
LGTM.

PR #261, opened 2021-05-25
Release 0.7.0

blootsvoets:
Changes since version 0.6.1:
- Separated Java project into smaller modules
  - Introduces radar-catalog-server application that replaces the catalog server from radar-schemas-tools.
  - Introduces radar-schemas-core library that replaces parsing and validation code from radar-schemas-tools.
  - Introduces radar-schemas-registration library that replaces topic and schema registration code from radar-schemas-tools.
- Fixed topic initialisation classpath and dependency issues
- Improved catalog-server logging.
- Read Avro schemas in via parser instead of via an avro tools Java class preprocessor.
- Added Github Actions


commons/push/garmin/garmin_pulse_ox.avsc
```diff
@@ -10,7 +10,7 @@
     {"name": "startTimeOffset", "type": ["null", "int"], "doc": "Offset in seconds to add to time to derive the 'local' time of the device that captured the data.", "default": null},
     {"name": "date", "type": ["null", "string"], "doc": "The calendar date this summary would be displayed on in Garmin Connect in the local time zone of the device. The date format is ‘yyyy-mm-dd’.", "default": null},
     {"name": "duration", "type": ["null", "int"], "doc": "Length of the monitoring period in seconds. 86400 once a full day is complete, but less if a user syncs mid-day.", "default": null},
-    {"name": "spo2Value", "type": ["null", "float"], "doc": "The SpO2 measurement taken at the time (1 sample/minute)", "default": null},
+    {"name": "spo2", "type": ["null", "float"], "doc": "The SpO2 measurement taken at the time (1 sample/minute).", "default": null},
```
mpgxvii, line 13: I think we will need to update this in `RADAR-PushEndpoint` as well. I can make the change there when needed.

mpgxvii (approved):
LGTM

commons/push/garmin/garmin_pulse_ox.avsc
blootsvoets: Thanks, forgot about that

yatharthranjan: Hi @mpgxvii can we not update the endpoint with new schema? The data collection is already ongoing and the data is being aggregated and put in postgres and then used for real-time analysis. This would require changes in multiple places and will create inconsistencies in the data and is not worth it IMO.

mpgxvii: Okay sure 👍

mpgxvii: I think we can maybe rollback this change since the data with the previous schema is already being used @yatharthranjan @blootsvoets?

yatharthranjan: Yes, IMO makes sense to keep it consistent if the change is not being included in the push-endpoint.

blootsvoets: You may need to change a validator rule though -- the validate task now fails if a field ends with `Value`.

blootsvoets: Would it be possible for the next release of the Garmin connector use radar-schemas-commons 0.7.0 with the new value with some compatibility in the streams? That way we don't have to keep supporting deprecated schemas forever. On any server installation, you can locally use the old schema with radar-schemas-tools init by overriding the docker path `/schema/conf/commons/push/garmin/garmin_pulse_ox.avsc` with the old value. When you're ready to update the garmin push connector, you can:

1. stop the push connector
2. stop the streams after it has processed the push values still in the pipeline OR write the stream so it is backwards compatible
3. deploy the topic_init.sh with the new schema
4. deploy the updated push connector
5. deploy the updated stream processor

yatharthranjan: Hi, it is not just the streams (KSQL) but multiple other downstream apps (like postgres for realtime models, timescale for grafana) also the extracted data too which will now be inconsistent with the data before the update.
All this and the scale of changes required make this minor change a major migration effort. So, if possible would be great to revert it to `spo2Value`. Maybe the validator can be configured to ignore this validation in the push/garmin sources?
Otherwise, we can update it here, and do not update the push-endpoint till the EOL of the KCL deployment (sometime in 2022).

blootsvoets: I checked the code and see that I already removed this check, anticipating this change. It's been a while ago. I'll revert this change due for 0.7.1.

PR #269, opened 2021-07-23
Improve retry on topic creation

baixiac:
The starting up of the `radar-schemas-tools` docker image could fail after the "double-checked" topic creation due to unavailable brokers. This PR tries to improve the retry logic.

### DIFF
--- a/docker/topic_init.sh
+++ b/docker/topic_init.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+MAX_RETRIES=5
+
 if [ -z NO_VALIDATE ]; then
   radar-schemas-tools validate merged
 fi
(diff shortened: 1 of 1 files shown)


docker/topic_init.sh
```diff
@@ -1,23 +1,29 @@
...
+    if radar-schemas-tools create -c "${KAFKA_CONFIG_PATH}" -p $KAFKA_NUM_PARTITIONS -r $KAFKA_NUM_REPLICATION -b $KAFKA_NUM_BROKERS -s "${KAFKA_BOOTSTRAP_SERVERS}" merged; then
+        echo "Created topics at attempt $i"
+        break
+    else
+        if [ $i -eq $MAX_RETRIES ]; then
+            echo "FAILED TO CREATE TOPICS AFTER $i ATTEMPT(S)"
+            exit 1
+        else
+            echo "FAILED TO CREATE TOPICS ... Retrying"
+            sleep 10
```
yatharthranjan, line 22: there is sleep logic within the application too so you can retry immediately too. Alternatively, add `Retrying after waiting for 10 seconds`in the echo to keep users informed that it will wait.

yatharthranjan (approved):
LGTM

docker/topic_init.sh
baixiac: Oh good to know. Sleep removed.

PR #270, opened 2021-07-26
Release 0.7.2

yatharthranjan:
- Improves retry logic for topic creation
- Additional questionnaires - Pulse Ox response, Unscripted Audio 4 (contains the whole North wind text in a single questionnaire), Audio Count, Audio Vocalisation

### DIFF
--- a/docker/topic_init.sh
+++ b/docker/topic_init.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+MAX_RETRIES=5
+
 if [ -z NO_VALIDATE ]; then
   radar-schemas-tools validate merged
 fi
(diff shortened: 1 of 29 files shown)


docker/topic_init.sh
```diff
@@ -1,23 +1,28 @@
 #!/bin/bash
 
+MAX_RETRIES=5
```
blootsvoets, line 3: Please make this variable accept environment variable, e.g.
```suggestion
MAX_RETRIES=${TOPIC_INIT_RETRIES:-5}
```

blootsvoets:
I'd prefer a solution where an additional `KafkaTopics.initialize(int numBrokers, int retries)` function is created, and the number of retries can be specified on the command line. If possible, I'd like to minimise the amount of bash code needed for this.

yatharthranjan:
ok sure, will make the changes

java-sdk/radar-schemas-tools/src/main/java/org/radarbase/schema/tools/SubCommand.java
```diff
@@ -39,4 +43,30 @@ static void addRootArgument(ArgumentParser parser) {
...
+    class IntRangeArgumentChoice implements ArgumentChoice {
+
+        private final int minRange;
+        private final int maxRange;
+        private final Set<Integer> range;
+
+        public IntRangeArgumentChoice(int minRange, int maxRange) {
+            this.minRange = minRange;
+            this.maxRange = maxRange;
+            this.range = IntStream
```
blootsvoets, line 56: This uses relatively much memory. Would propose to change the contains check to
```java
val instanceof Integer && val >= minRange && val <= maxRange;
```

blootsvoets (approved):
LGTM. Small comment but not blocking.

PR #274, opened 2021-08-04
add new questionnaires to specs

yatharthranjan:
- K BUILD Private
- NuvoAir Spirometry Notifications
- Exacerbation diary for COPD

### DIFF
--- a/specifications/active/aRMT-1.15.0.yml
+++ b/specifications/active/aRMT-1.15.0.yml
@@ -1,0 +1,263 @@
+name: aRMT
+vendor: RADAR
+model: aRMT-App
+version: 1.15.0
+assessment_type: QUESTIONNAIRE
+doc: aRMT Questionnaires definition. Includes Personal Health Questionnaire Depression Scale (PHQ-8), Experience sampling method (ESM) and RSES and other data.
+data:
+  - type: THINC_IT
+    topic: notification_thinc_it
+    value_schema: .active.questionnaire.Questionnaire
+    questionnaire_definition_url: https://github.com/RADAR-base/RADAR-REDCap-aRMT-Definitions/blob/master/questionnaires/thinc_it/thinc_it_armt.json
+  - type: ROMBERG_TEST
+    doc: The value of the task is in milliseconds (ms).
+    topic: task_romberg_test
+    value_schema: .active.task.Task
+    questionnaire_definition_url: https://github.com/RADAR-base/RADAR-REDCap-aRMT-Definitions/blob/master/questionnaires/romberg_test/romberg_test_armt.json
+  - type: 2MW_TEST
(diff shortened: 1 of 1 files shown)


specifications/active/aRMT-1.15.0.yml
```diff
@@ -0,0 +1,263 @@
...
+    questionnaire_definition_url: https://raw.githubusercontent.com/RADAR-base/RADAR-REDCap-aRMT-Definitions/master/questionnaires/epi_seizure_diary/epi_seizure_diary_armt.json
+  - type: EPI_EVENING_QUESTIONNAIRE
+    topic: questionnaire_epi_evening_questionnaire
+    value_schema: .active.questionnaire.Questionnaire
+    questionnaire_definition_url: https://raw.githubusercontent.com/RADAR-base/RADAR-REDCap-aRMT-Definitions/master/questionnaires/epi_evening_questionnaire/epi_evening_questionnaire_armt.json
+  - type: EPI_WSAS
+    topic: questionnaire_epi_wsas
+    value_schema: .active.questionnaire.Questionnaire
+    questionnaire_definition_url: https://raw.githubusercontent.com/RADAR-base/RADAR-REDCap-aRMT-Definitions/master/questionnaires/epi_wsas/epi_wsas_armt.json
+  - type: BP_QUESTIONNAIRE
```
mpgxvii, line 256: This is for another study but just realized yesterday from the protocol, the name is `bp`. So it might be better to change this to `type: BP`. What do you think @nivemaham?

specifications/active/aRMT-1.15.0.yml
```diff
@@ -0,0 +1,263 @@
...
+    topic: questionnaire_bp
+    value_schema: .active.questionnaire.Questionnaire
+    questionnaire_definition_url: https://raw.githubusercontent.com/RADAR-base/RADAR-REDCap-aRMT-Definitions/master/questionnaires/bp/bp_armt.json
+  - type: HLH_QUESTIONNAIRE
```
mpgxvii, line 260: Same here

specifications/active/aRMT-1.15.0.yml
yatharthranjan: Is the type used anywhere? IIRC it is not, so we can change it as it will have no implications and it is nice to keep these consistent.

mpgxvii: Yes it's used by the `HLH-Pilot` study so I'm not sure if we can change.

specifications/active/aRMT-1.15.0.yml
nivemaham: Sounds good

mpgxvii (approved):
LGTM

PR #276, opened 2021-08-12
YAML specification added for DREEM device

farazkhalidamani:
Hi,

To provide support for DREEM devices a new YAML specification has been created.
It has all relevant information related to the Schema, Topic, Doc etc.

Thank you.

Regards,
Faraz


specifications/passive/dreem-1.0.0.yml
```diff
@@ -0,0 +1,21 @@
...
+    unit: NON_DIMENSIONAL
+    processing_state: VENDOR
+    topic: dreem_sleep_stage
+    value_schema: .passive.dreem.DreemSleepStage
+  - type: DREEM SLEEP ENDPOINTS
+    unit: NON_DIMENSIONAL
+    processing_state: VENDOR
+    topic: dreem_sleep_endpoints
+    value_schema: .passive.dreem.DreemSleepEndpoints
+  - type: DREEM METRIC
```
nivemaham, line 17: I think there are only two data topics. One is sleep endpoints and the other is the sleep stage
The Dreem metric is used within the sleep endpoints.

specifications/passive/dreem-1.0.0.yml
```diff
@@ -0,0 +1,21 @@
+#====================================== For DREEM device =====================================#
+vendor: DREEM
+model: NA
```
nivemaham, line 3: Isn't there a model for the device?

specifications/passive/dreem-1.0.0.yml
```diff
@@ -0,0 +1,21 @@
...
+version: 1.0.0
+doc: DREEM data collected and uploaded by DREEM web application
+data:
+  - type: DREEM SLEEP STAGE
```
nivemaham, line 7: I think the type property here is not relevant, thus can be removed. Other wise can you follow the syntax from other specs and append the works with an `_` and remove the DREEN prefix?

specifications/passive/dreem-1.0.0.yml
```diff
@@ -0,0 +1,14 @@
+#====================================== For Dreem 3 headband also known as Neuroband device =====================================#
+vendor: DREEM
+model: Dreem 3 headband also known as Neuroband
+version: 1.0.0
+doc: DREEM data collected and uploaded by DREEM web application
+data:
+  - unit: NON_DIMENSIONAL
```
nivemaham, line 7: Can you also add a `doc` property with the description of the data ?
example
```yaml
  - doc: Data from 3-axis accelerometer sensor with gravitational constant g as unit. Data extracted from CWA binary file.
    topic: connect_upload_axivity_acceleration
    value_schema: .connector.upload.axivity.AxivityAcceleration
```

nivemaham (approved):
LGTM

Issue #272, opened 2021-07-30
The new log obfuscates the aggregation on Kafka topics


baixiac:
Looks like the logging in the new release confuses `radar-docker install` when it tries to aggregate topics by getting the last line from stdout of list_aggregated.sh (using command `radar-schemas-tools`):
```
radar-schema-tools is initialized with root directory /schema/merged [org.radarbase.schema.tools.CommandLineApp:63] ,android_biovotion_vsm1_acceleration_10min,android_biovotion_vsm1_acceleration_10sec,android_biovotion_vsm1_acceleration_1day,android_biovotion_vsm1_acceleration_1hour,android_biovotion_vsm1_acceleration_1min,android_biovotion_vsm1_acceleration_1week...
```
rather than expected:
```
android_biovotion_vsm1_acceleration_10min,android_biovotion_vsm1_acceleration_10sec,android_biovotion_vsm1_acceleration_1day,android_biovotion_vsm1_acceleration_1hour,android_biovotion_vsm1_acceleration_1min,android_biovotion_vsm1_acceleration_1week...
```
Thus, `./etc/mongodb-connector/sink-mongo.properties` becomes invalid:
```
...
topics=radar-schema-tools is initialized with root directory /schema/merged [org.radarbase.schema.tools.CommandLineApp:63] ,android_biovotion_vsm1_acceleration_10min,android_biovotion_vsm1_acceleration_10sec,android_biovotion_vsm1_acceleration_1day,android_biovotion_vsm1_acceleration_1hour,android_biovotion_vsm1_acceleration_1min,android_biovotion_vsm1_acceleration_1week...
```

Here is the call site of the logging:
https://github.com/RADAR-base/RADAR-Schemas/blob/6d519df65f4ff021580bbf846392c1a38f9f0617/java-sdk/radar-schemas-tools/src/main/java/org/radarbase/schema/tools/CommandLineApp.java#L63

baixiac:
The maximum no. of retries on refreshing topics has been set to 10 and that's oh. However, the retry back-off seems not working/existing or the timeout is too short so they all happed within one sec:
https://github.com/RADAR-base/RADAR-Docker/runs/3203037035?check_suite_focus=true#step:6:185

More details in here:
https://github.com/RADAR-base/RADAR-Schemas/blob/master/java-sdk/radar-schemas-registration/src/main/java/org/radarbase/schema/registration/KafkaTopics.java#L193-L230

[...]

PR #280, opened 2021-08-23
Removes log message when using quiet flag

blootsvoets:
Log messages are no longer displayed if the quiet flag is given. Fixes part of #272.

Note: gradle update is already applied in dev branch.


yatharthranjan (approved):
LGTM

PR #282, opened 2021-08-24
Unify topic listing with broker listing

### LINKED ISSUE #281: Retry sleep/backoff not working
moved from #272 

> The maximum no. of retries on refreshing topics has been set to 10 and that's oh. However, the retry back-off seems not working/existing or the timeout is too short so they all happed within one sec:
https://github.com/RADAR-base/RADAR-Docker/runs/3203037035?check_suite_focus=true#step:6:185

> More details in here:
https://github.com/RADAR-base/RADAR-Schemas/blob/master/java-sdk/radar-schemas-registration/src/main/java/org/radarbase/schema/registration/KafkaTopics.java#L193-L230

blootsvoets:
This PR represents a more consistent approach to trying and retrying, duplicating the logic between checking for brokers and listing topics. Fixes #281.


java-sdk/radar-schemas-registration/src/main/java/org/radarbase/schema/registration/KafkaTopics.java
```diff
@@ -193,40 +204,42 @@ public boolean createTopics(Stream<String> topicsToCreate, int partitions, short
...
-                logger.warn("Topics not listed yet after {} seconds", sleep);
+            } else if (tries < numTries - 1) {
+                logger.warn("Topics not listed yet. Trying again after {} seconds.", sleep);
+                long timeTaken = System.currentTimeMillis() - start;
+                long sleepMillis = Math.max(sleep * 1000L - timeTaken, 0L);
+                if (sleepMillis > 0L) {
+                    Thread.sleep(sleepMillis);
+                }
+                sleep = Math.min(MAX_SLEEP, sleep * 2);
+                timeout = Math.min(5, sleep);
```
yatharthranjan, line 237: can we just use sleep or is this min necessary?

java-sdk/radar-schemas-registration/src/main/java/org/radarbase/schema/registration/KafkaTopics.java
```diff
@@ -193,40 +204,42 @@ public boolean createTopics(Stream<String> topicsToCreate, int partitions, short
...
-            } catch (TimeoutException e) {
-                // do nothing
+                topics = getKafkaClient()
+                        .listTopics(opts)
+                        .names()
+                        .get(sleep, TimeUnit.SECONDS);
+            } catch (ExecutionException ex) {
+                logger.error("Failed to list topics from brokers: {}", ex.getCause().toString());
+            } catch (TimeoutException ex) {
+                logger.error("Failed to list topics within {} seconds", timeout);
```
yatharthranjan, line 225: can this be sleep? Since that is the time specified in the get call. Otherwise the get call should also use timeout?

yatharthranjan (approved):
LGTM

java-sdk/radar-schemas-registration/src/main/java/org/radarbase/schema/registration/KafkaTopics.java
blootsvoets: Indeed, done.

java-sdk/radar-schemas-registration/src/main/java/org/radarbase/schema/registration/KafkaTopics.java
blootsvoets: Indeed, done.

PR #285, opened 2021-09-06
Update altoida_summary schema

mpgxvii:
- Update `altoida_summary` schema to allow nullable fields


commons/connector/upload/altoida/altoida_summary.avsc
```diff
@@ -7,13 +7,13 @@
...
-    {"name": "yearsOfEducation", "type": "int", "doc": "Years of education in years."},
-    {"name": "gender", "type": {
+    {"name": "age", "type": ["null","int"], "doc": "Age of the participant in years.",  "default": null},
+    {"name": "yearsOfEducation", "type": ["null","int"], "doc": "Years of education in years.",  "default": null},
+    {"name": "gender", "type": ["null", {
           "name": "GenderType",
           "type": "enum",
           "doc": "Gender types provided by Altoida.",
-          "symbols": ["MALE", "FEMALE", "OTHER", "UNKNOWN"]}, "doc": "Gender of the participant.", "default": "UNKNOWN"},
+          "symbols": ["MALE", "FEMALE", "OTHER", "UNKNOWN"]}], "doc": "Gender of the participant.", "default": null},
```
blootsvoets, line 16: The `UNKNOWN` symbol is meant to replace `null`, so you could store `UNKNOWN` whenever the gender is not passed in the data. From the contributing guidelines:
> - For fields that are inherent to a record, and will never be removed or renamed, no default value is needed. For all other fields:
>   - if the type is an enum, use an `UNKNOWN` symbol as default value
>   - otherwise, set the type to a union of `["null", <intended type>]` and set the default value to `null`.

blootsvoets (approved):
LGTM

PR #287, opened 2021-09-23
add exacerbation diary ild to specs

yatharthranjan:


mpgxvii:
@yatharthranjan @afolarin Do you know if we should already add the Vaping study questionnaire to the schema specs?

yatharthranjan:
Not sure, but can keep this open to be added to this later? Not urgent to merge this as have already added the schema and topic to the server

PR #293, opened 2021-12-14
Release 0.7.5

blootsvoets:
Changes since 0.7.4:
- **Critical**: fixes Log4J CVE-2021-44228 in catalog-server
- Updates other dependencies, including Gradle 7.3 and Java 17
- Adds CANEdge schemas.
- Adds check that specifications can be parsed.

### DIFF
--- a/java-sdk/radar-schemas-core/src/main/java/org/radarbase/schema/specification/active/questionnaire/QuestionnaireDataTopic.java
+++ b/java-sdk/radar-schemas-core/src/main/java/org/radarbase/schema/specification/active/questionnaire/QuestionnaireDataTopic.java
@@ -28,7 +28,7 @@ import org.radarbase.schema.specification.DataTopic;
  */
 @JsonInclude(Include.NON_NULL)
 public class QuestionnaireDataTopic extends DataTopic {
-    @JsonProperty
+    @JsonProperty("questionnaire_definition_url")
     private URL questionnaireDefinitionUrl;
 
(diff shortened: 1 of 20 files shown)


java-sdk/radar-schemas-core/src/main/java/org/radarbase/schema/specification/active/questionnaire/QuestionnaireDataTopic.java
```diff
@@ -38,6 +38,6 @@ public URL getQuestionnaireDefinitionUrl() {
     @Override
     protected void propertiesMap(Map<String, Object> props, boolean reduced) {
         super.propertiesMap(props, reduced);
-        props.put("questionnaireDefinitionUrl", questionnaireDefinitionUrl);
+        props.put("questionnaire_definition_url", questionnaireDefinitionUrl);
```
nivemaham, line 41: do we have to write the specifications for adding new questionnaires after using this version? @mpgxvii

nivemaham (approved):
LGTM

java-sdk/radar-schemas-core/src/main/java/org/radarbase/schema/specification/active/questionnaire/QuestionnaireDataTopic.java
blootsvoets: I checked the questionnaire code and this variable is not used. The new definition mimics the style of the other variables.

PR #301, opened 2022-03-23
Add new resource fields restingHeartRate and sleep efficiency

lalitagarwal:
This PR is related to the issue [#89](https://github.com/RADAR-base/RADAR-REST-Connector/issues/89)

Changes made in the PR:
- Add `restingHeartRate` field to the intraday heart rate schema
- Add `efficiency` field to the sleep classic and stage schemas

### DIFF
--- a/commons/connector/fitbit/fitbit_intraday_heart_rate.avsc
+++ b/commons/connector/fitbit/fitbit_intraday_heart_rate.avsc
@@ -8,6 +8,6 @@
     { "name": "timeReceived", "type": "double", "doc": "Time that the data was received from the Fitbit API (seconds since the Unix Epoch)." },
     { "name": "timeInterval", "type": "int", "doc":  "Chronological window size (s)." },
     { "name": "heartRate", "type": "int", "doc": "Heart rate value (bpm)."},
-    { "name": "restingHeartRate", "type": "int", "doc": "Resting heart rate value for the day. A sleep stage log is required to generate this value. When a classic sleep log is recorded, this value will be missing." }
+    { "name": "restingHeartRate", "type": ["null", "int"], "doc": "Resting heart rate value for the day. A sleep stage log is required to generate this value. When a classic sleep log is recorded, this value will be missing.", "default": null }
   ]
 }
--- a/commons/connector/fitbit/fitbit_intraday_heart_rate.avsc
+++ b/commons/connector/fitbit/fitbit_intraday_heart_rate.avsc
@@ -8,6 +8,6 @@
     { "name": "timeReceived", "type": "double", "doc": "Time that the data was received from the Fitbit API (seconds since the Unix Epoch)." },
     { "name": "timeInterval", "type": "int", "doc":  "Chronological window size (s)." },
     { "name": "heartRate", "type": "int", "doc": "Heart rate value (bpm)."},
-    { "name": "restingHeartRate", "type": ["null", "int"], "doc": "Resting heart rate value for the day. A sleep stage log is required to generate this value. When a classic sleep log is recorded, this value will be missing.", "default": null }
+    { "name": "restingHeartRate", "type": ["null", "int"], "doc": "Resting heart rate value (bpm) for the day. A sleep stage log is required to generate this value. When a classic sleep log is recorded, this value will be missing.", "default": null }
   ]
 }
--- a/commons/connector/fitbit/fitbit_intraday_heart_rate.avsc
+++ b/commons/connector/fitbit/fitbit_intraday_heart_rate.avsc
@@ -7,6 +7,7 @@
     { "name": "time", "type": "double", "doc": "Device timestamp in UTC (s)." },
     { "name": "timeReceived", "type": "double", "doc": "Time that the data was received from the Fitbit API (seconds since the Unix Epoch)." },
     { "name": "timeInterval", "type": "int", "doc":  "Chronological window size (s)." },
-    { "name": "heartRate", "type": "int", "doc": "Heart rate value (bpm)."}
+    { "name": "heartRate", "type": "int", "doc": "Heart rate value (bpm)."},
+    { "name": "restingHeartRate", "type": "int", "doc": "Resting heart rate value for the day. A sleep stage log is required to generate this value. When a classic sleep log is recorded, this value will be missing." }
   ]
 }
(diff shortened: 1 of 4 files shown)


commons/connector/fitbit/fitbit_sleep_classic.avsc
```diff
@@ -9,6 +9,7 @@
     { "name": "duration", "type": "int", "doc": "Duration at this sleep characteristic in seconds." },
     { "name": "level", "type":
       { "name": "FitbitSleepClassicLevel", "type": "enum", "symbols": ["AWAKE", "RESTLESS", "ASLEEP", "UNKNOWN"], "doc": "Level of sleep as computed by Fitbit."},
-      "doc": "Level of sleep as computed by Fitbit.", "default": "UNKNOWN" }
+      "doc": "Level of sleep as computed by Fitbit.", "default": "UNKNOWN" },
+    { "name": "efficiency", "type": "string", "doc": "Calculated sleep efficiency score" }
```
blootsvoets, line 13: Should be nullable, as above. Ideally, the efficiency is a number (int / float) and the doc string explains how it is calculated.

commons/connector/fitbit/fitbit_sleep_stage.avsc
```diff
@@ -9,6 +9,7 @@
     { "name": "duration", "type": "int", "doc": "Duration at this sleep characteristic in seconds." },
     { "name": "level", "type":
       {"name": "FitbitSleepStageLevel", "type": "enum", "symbols": ["DEEP", "LIGHT", "REM", "AWAKE", "UNKNOWN"], "doc": "Level of sleep as computed by Fitbit."},
-      "doc": "Level of sleep as computed by Fitbit.", "default": "UNKNOWN" }
+      "doc": "Level of sleep as computed by Fitbit.", "default": "UNKNOWN" },
+    { "name": "efficiency", "type": "string", "doc": "Calculated sleep efficiency score" }
```
blootsvoets, line 13: Should be nullable, as above. Ideally, the efficiency is a number (int / float) and the doc string explains how it is calculated.

commons/connector/fitbit/fitbit_intraday_heart_rate.avsc
```diff
@@ -7,6 +7,7 @@
     { "name": "time", "type": "double", "doc": "Device timestamp in UTC (s)." },
     { "name": "timeReceived", "type": "double", "doc": "Time that the data was received from the Fitbit API (seconds since the Unix Epoch)." },
     { "name": "timeInterval", "type": "int", "doc":  "Chronological window size (s)." },
-    { "name": "heartRate", "type": "int", "doc": "Heart rate value (bpm)."}
+    { "name": "heartRate", "type": "int", "doc": "Heart rate value (bpm)."},
+    { "name": "restingHeartRate", "type": "int", "doc": "Resting heart rate value for the day. A sleep stage log is required to generate this value. When a classic sleep log is recorded, this value will be missing." }
```
blootsvoets, line 11: Isn't this generated in the summary instead of part of the intraday heart rate? In any case, whenever a new field is added, please make it nullable as follows: `"type": ["null", "<actual type>"], "default": null`. This reduces incompatibilities with existing data.

blootsvoets (changes requested):
Thanks for your contribution, this looks good, just some small comments.

commons/connector/fitbit/fitbit_sleep_classic.avsc
lalitagarwal: Thanks for pointing this out. I have changed the type to `int` and added null as a default value.

commons/connector/fitbit/fitbit_intraday_heart_rate.avsc
lalitagarwal: I notice that the restingHeartRate field is present in both the intraday response and the summary response. Correct me if am wrong. Should I also add this field to `fitbit_activity_heart_rate.avsc`?
- https://dev.fitbit.com/build/reference/web-api/heartrate-timeseries/get-heartrate-timeseries-by-date/
- https://dev.fitbit.com/build/reference/web-api/intraday/get-heartrate-intraday-by-date/

blootsvoets: Yes, it makes sense to add it to the activity heart rate as well.

blootsvoets:
Thanks for the changes. If possible, could you still add units (bpm, %) to the docs, and for sleep `efficiency` a more complete description of what this calculation means?

blootsvoets (approved):
Thanks for the swift updates! Looks good to me!

yatharthranjan:
Hi, sorry for joining a bit late as was on sick leave last week.

This looks good to me, however, will have a lot of data redundancy as the restingHeartRate will be included for each intraday data point (however it is only specified daily by fitbit)

Some considerations for the resting heart rate- 
the restingHeartRate provided as part of the intraday time series is for daily data (at the top of the records there is data per date for this summary), as such it does not make sense to include it with each intraday heart rate record. I would propose to have a separate route to get this summary only information from the non-intraday endpoint.
An alternative would be to create this as a separate Record when processing the intraday heart rate (since we are already getting this info) and only create the record once for each date. In both cases, this will need to be in a separate topic and schema.

what do you think @blootsvoets

blootsvoets:
Yes, that makes sense. I wasn't aware that this data was static per day. OK then this change should be reverted indeed. If there's another API endpoint for this information that would make sense. Making it a separate call would also helps users that do not have intraday access.

yatharthranjan:
Hi @lalitagarwal, Thanks for the contribution. Based on the conversation, can you please remove the restingHeartRate from the `fitbit_intraday_heart_rate.avsc` and create a new schema `fitbit_heart_rate_timeseries.avsc` based on [this API](https://dev.fitbit.com/build/reference/web-api/heartrate-timeseries/get-heartrate-timeseries-by-date-range/). You can leave out the `customHeartRateZones` and `heartRateZones` for now as they are not needed. We can add later if required. So it will be just `dateTime` and `restingHeartRate` along with the required radar fields. I am reverting this merge now. 
Thank you.

lalitagarwal:
Thanks for the comments @yatharthranjan and @blootsvoets. I have created a new PR with the proposed changes. Please take a look once possible: https://github.com/RADAR-base/RADAR-Schemas/pull/303

PR #304, opened 2022-04-05
Make docker login optional

blootsvoets:
Github actions will no longer do a docker login for external contributors.
Additionally, build all docker images for all platforms, also in `dev`.


yatharthranjan (approved):
LGTM.

PR #309, opened 2022-07-06
Add health schema

hardco2020:
# Discussion

## How do we define the schema ?

1. Create a schema for each type ( Bad Idea )

2. Aggregate some of the similar types so we don't have to create so many schemas, if that's the case, how should we make the aggregation?

3. Other great ideas we haven't thought about!!


commons/connector/health/health_activity.avsc
```diff
@@ -0,0 +1,24 @@
...
+    },
+    {
+      "name": "timeInterval",
+      "type": "int",
+      "doc": "Chronological window size (s)."
+    },
+    {
+      "name": "activity",
+      "type": "int",
+      "doc": "Activity taken in this period."
```
mpgxvii, line 21: Is there a defined set of possible activities that can be taken? If so, we can maybe add them as enums, similar to: https://github.com/RADAR-base/RADAR-Schemas/blob/master/commons/monitor/questionnaire/questionnaire_application_interaction_event.avsc#L12

commons/connector/health/health_appleExerciseTime.avsc
```diff
@@ -0,0 +1,24 @@
+{
+  "namespace": "org.radarcns.connector.health",
+  "name": "HKQuantityTypeIdentifierAppleExerciseTime",
+  "type": "record",
+  "doc": "AppleExerciseTime , unit is min.",
```
mpgxvii, line 5: Does `apple` here stand for the Apple watch?

commons/connector/health/health_height.avsc
```diff
@@ -0,0 +1,15 @@
+{
+  "namespace": "org.radarcns.connector.health",
+  "name": "HKQuantityTypeIdentifierHeight",
```
mpgxvii, line 3: I think this can be combined with other body measurements as one schema. Maybe called `HKQuantityTypeIdentifierBodyMeasurements`. This can be combined with height, body mass, body mass index, lean body mass, body fat percentage, and weight circumference. As categorised here: https://developer.apple.com/documentation/healthkit/

commons/connector/health/health_heartRate.avsc
```diff
@@ -0,0 +1,24 @@
+{
+  "namespace": "org.radarcns.connector.health",
+  "name": "HKQuantityTypeIdentifierHeartRate",
```
mpgxvii, line 3: This can also be combined with other Vital Signs, as categorised here: https://developer.apple.com/documentation/healthkit/
This can be combined with heart rate variability, resting heart rate, etc.

commons/connector/health/health_activity.avsc
```diff
@@ -0,0 +1,24 @@
+{
+  "namespace": "org.radarcns.connector.health",
```
blootsvoets, line 2: If this is a HealthKit connector, please rename the namespace (and directory) to `org.radarcns.connector.healthkit`. If it is meant to be integrated into the aRMT app, please rename (and move) to `org.radarcns.active.healthkit`

commons/connector/health/health_activity.avsc
```diff
@@ -0,0 +1,24 @@
...
+  "name": "HKWorkoutTypeIdentifier_HKCategoryTypeIdentifierSleepAnalysis",
```
blootsvoets, line 3: The package identifies that this is about HealthKit, the fact that this is a schema gives it a type identifier. So adding `TypeIdentifier` to the name is redundant. Alternative proposal: `HealthKitActivity`

commons/connector/health/health_activity.avsc
```diff
@@ -0,0 +1,24 @@
...
+    {
+      "name": "timeInterval",
+      "type": "int",
+      "doc": "Chronological window size (s)."
+    },
+    {
+      "name": "activity",
+      "type": "int",
+      "doc": "Activity taken in this period."
+    }
```
blootsvoets, line 22: Please use an [Avro enum](https://avro.apache.org/docs/current/spec.html#Enums) here to list the possible activities. First in the list can be `UNKNOWN`, and please also include `OTHER`, where `UNKNOWN` mens you do not know what the activity was, and `OTHER` means the activity was not included in the enum.

commons/connector/health/health_appleExerciseTime.avsc
```diff
@@ -0,0 +1,24 @@
...
+      "type": "double",
+      "doc": "Time that the data was received from the Apple Healthkit ."
+    },
+    {
+      "name": "timeInterval",
+      "type": "int",
+      "doc": "Chronological window size (s)."
+    },
+    {
+      "name": "appleExerciseTime",
```
blootsvoets, line 19: The package identifies that this is about Apple HealthKit. This can be renamed to `exerciseDuration` to avoid confusion with a timestamp. In the field doc, you could document what Apple HealthKit variable this exerciseDuration corresponds to, e.g. HKQuantityTypeIdentifierAppleExerciseTime.

commons/connector/health/health_appleExerciseTime.avsc
```diff
@@ -0,0 +1,24 @@
...
+      "type": "int",
+      "doc": "Exercise time  taken in this period."
```
blootsvoets, line 21: Please add a unit to the docs (minutes / seconds).

commons/connector/health/health_appleExerciseTime.avsc
```diff
@@ -0,0 +1,24 @@
+{
+  "namespace": "org.radarcns.connector.health",
+  "name": "HKQuantityTypeIdentifierAppleExerciseTime",
```
blootsvoets, line 3: This can be just HealthKitExerciseTime

commons/connector/health/health_appleExerciseTime.avsc
blootsvoets: In the record doc, there's no need to add the unit, please add it to the field doc instead. In the record doc, you can document how these records are obtained (i.e. via Apple HealthKit).

blootsvoets (commented):
Thanks for your contribution! I've listed some comments in the first schemas about how the naming and documentation could be improved. Please consider those comments for the other schemas as well.

blootsvoets (changes requested):
This still needs some work before merging, please see my previous comments.

Records that might be merged:
- glucose / insulin measurements
- calories
- blood pressure / heart rate measurements
- height / weight / bmi / waist
- nutritional values

I might be missing some. For content fields of these merged records, please use `"type": ["null", "<type>"], "default": null` so that not all values need to be filled (e.g. if the sensor does not carry the value, it has not been entered in HealthKit yet, or the user has not given permission for that particular value).

hardco2020:
@blootsvoets @mpgxvii  I made some changes based on your suggestions.

There are two sample aggregation files  `health_mergingExample1.avsc`, `health_MergingExample2.avsc` since I am not entirely sure how the format of aggregation will be, so I created two files for you guys to review, let me know if there's more details needed to be revise and which file is close to your expectation for aggregation schema.

commons/connector/health/health_mergingExample1.avsc
```diff
@@ -0,0 +1,62 @@
...
+        "name": "time",
+        "type": "double",
+        "doc": "Device timestamp in UTC (s)."
+      },
+      {
+        "name": "timeReceived",
+        "type": "double",
+        "doc": "Time that the data was received from the Apple Healthkit ."
+      },
+      { "name": "height", "type": "int", "doc": "Users's height, unit is m." }
```
mpgxvii, line 19: Thanks for updating this. I think the aggregation that we require would be for the actual fields. For example:
```{
   "namespace":"org.radarcns.active.healthkit",
   "name":"HealthKitBodyMeasurements",
   "type":[
      "null",
      "record"
   ],
   "default":null,
   "doc":"Health Kit body measurements..",
   "fields":[
      {
         "name":"time",
         "type":"double",
         "doc":"Device timestamp in UTC (s)."
      },
      {
         "name":"timeReceived",
         "type":"double",
         "doc":"Time that the data was received from the Apple Healthkit ."
      },
      {
         "name":"height",
         "type":"int",
         "doc":"Users's height, unit is m."
      },
      {
         "name":"weight",
         "type":"int",
         "doc":"Users's weight, unit is kg."
      },
      {
         "name":"bmi",
         "type":"int",
         "doc":"Users's bmi, unit is count."
      }
   ]
}.....

File: commons/connector/health/health_mergingExample2.avsc [hunk 12]
```diff
@@ -0,0 +1,48 @@
+{
+  "namespace": "org.radarcns.active.healthkit",
+  "name": "HealthKit_Height_BodyMass_BodyMassIndex",
```
mpgxvii, line 3: My bad, didn't see this. Yes it should be like this, but without the duplicate `time` and `timeReceived` fields.

commons/connector/health/aggregation/health_blood_measurements.avsc
```diff
@@ -0,0 +1,49 @@
+{
+  "namespace": "org.radarcns.active.healthkit",
+  "name": "HealthKitBloodMeasurements",
+  "type": ["null", "record"],
```
blootsvoets, line 4: In the record type, don't add `null`.

commons/connector/health/aggregation/health_blood_measurements.avsc
```diff
@@ -0,0 +1,49 @@
...
+      "type": "double",
+      "doc": "Device timestamp in UTC (s)"
+    },
+    {
+      "name": "timeReceived",
+      "type": "double",
+      "doc": "Time that the data was received from the Apple Healthkit"
+    },
+    {
+      "name": "blood_pressure_systolic",
```
blootsvoets, line 19: Use camel case (`bloodPressureSystolic`)

commons/connector/health/aggregation/health_blood_measurements.avsc
```diff
@@ -0,0 +1,49 @@
...
+    },
+    {
+      "name": "vo2max",
+      "type": "double",
+      "doc": "Users's vo2max, unit is ml/(kg * min)"
+    },
+    {
+      "name": "insulin",
+      "type": "double",
+      "doc": "User taken insulin, unit is IU"
```
blootsvoets, line 46: If there's a chance that some of these values are unknown, please make their type `["null", "double"]` instead and add `"default": null`.

commons/connector/health/aggregation/health_body_measurements.avsc
```diff
@@ -0,0 +1,65 @@
+{
+  "namespace": "org.radarcns.active.healthkit",
+  "name": "HealthKitBodyMeasurements",
+  "type": ["null", "record"],
+  "default": null,
+  "doc": "height, weight, bmi, fat_percentage, temperature, waist_circumference, gender, date_of_birth",
```
blootsvoets, line 6: The fields can be read from the field list. Please add a description of this record type instead.

commons/connector/health/aggregation/health_body_measurements.avsc
```diff
@@ -0,0 +1,65 @@
+{
+  "namespace": "org.radarcns.active.healthkit",
+  "name": "HealthKitBodyMeasurements",
```
blootsvoets, line 3: The record name would ideally be singular.

commons/connector/health/aggregation/health_body_measurements.avsc
```diff
@@ -0,0 +1,65 @@
...
+    {
+      "name": "waist_circumference",
+      "type": "double",
+      "doc": "Users's waist circumference, unit is m"
+    },
+    {
+      "name": "gender",
+      "type": "enum",
+      "doc": "User's gender"
+      "symbols" : ["male","female","other"]
```
blootsvoets, line 52: enum in avro needs to be a full type specification, i.e.
```suggestion
      "type": {
        "name": "HealthKitGender",
        "type": "enum",
        "doc": "Health kit gender options, unknown if not provided."
        "symbols" : ["UNKNOWN", "MALE","FEMALE","OTHER"]
      },
      "doc": "User's gender"
      "default": "UNKNOWN"
```

commons/connector/health/aggregation/health_body_measurements.avsc
```diff
@@ -0,0 +1,65 @@
...
+    },
+    {
+      "name": "date_of_birth",
+      "type": "",
+      "doc": "Users's date of birth"
```
blootsvoets, line 57: Please specify format in doc, preferably YYYY-MM-DD (as per ISO 8601).

commons/connector/health/aggregation/health_body_measurements.avsc
```diff
@@ -0,0 +1,65 @@
...
+    },
+    {
+      "name": "mindfulness",
+      "type": "int",
+      "doc": "Users's date of birth, unit is sec"
```
blootsvoets, line 62: Copy paste error?

commons/connector/health/aggregation/health_calories_measurements.avsc
```diff
@@ -0,0 +1,29 @@
...
+      "type": "double",
+      "doc": "Device timestamp in UTC (s)."
+    },
+    {
+      "name": "timeReceived",
+      "type": "double",
+      "doc": "Time that the data was received from the Apple Healthkit ."
+    },
+    {
+      "name": "calories.active",
```
blootsvoets, line 19: Please use camel case

commons/connector/health/aggregation/health_exercise_measurements.avsc
```diff
@@ -0,0 +1,46 @@
...
+      "doc": "Users's distance, unit is m"
+    },
+    {
+      "name": "appleExerciseTime",
+      "type": "double",
+      "doc": "Users's appleExerciseTime, unit is min"
+    },
+    {
+      "name": "activity",
+      "type": "enum",
```
blootsvoets, line 40: See earlier comment about enum

commons/connector/health/aggregation/health_nutrition_measurements.avsc
```diff
@@ -0,0 +1,114 @@
...
+      "doc": "Device timestamp in UTC (s)."
+    },
+    {
+      "name": "timeReceived",
+      "type": "double",
+      "doc": "Time that the data was received from the Apple Healthkit ."
+    },
+    {
+      "name": "nutrition.calories",
+      "type": "double",
```
blootsvoets, line 20: It's not necessarily bad to have these as doubles, but it may be sufficient to have them as `float` instead if you expect each value to have six or fewer digits. Same for all the values below.

blootsvoets (commented):
The aggregated schema's look alright, I've added some comments. When you have something you're happy with, please run
```shell
cd java-sdk
./gradlew test
```
to see if the Avro schema's compile and are valid (requires Java 11+ to be installed)

commons/connector/health/aggregation/health_blood_measurements.avsc
hardco2020: There are _ and . for naming because when we are getting the info from cordova-health, it will need this particular name to get the info, so If we want to use different value in our schema, I have to write a lot of  (if else) to judge the input, is there any constraints on schema to use _ or . ?

commons/connector/health/aggregation/health_body_measurements.avsc
hardco2020: Hi @blootsvoets  What do you mean by the field list, can you have an example what I should add here ?

commons/connector/health/aggregation/health_body_measurements.avsc
hardco2020: I discussed with @mpgxvii, we decided to go with string, something like "1998-07-28", then decode this info on front-end, let me know what you think

commons/connector/health/aggregation/health_calories_measurements.avsc
hardco2020: The same as above case, I have to specify it to be this way to get the data

PR #310, opened 2022-08-25
Add general aRMT topic

blootsvoets:
Adds a general `questionnaire_response` Kafka topic to the aRMT app, as well as moving the monitoring topics to that general device type. This allows an installation to have far fewer topics if the users do not require questionnaire responses to be separated over different topics. For example in Confluent Cloud this makes a big cost difference.

### DIFF
--- a/specifications/active/aRMT-1.17.0.yml
+++ b/specifications/active/aRMT-1.17.0.yml
@@ -1,0 +1,299 @@
+name: aRMT
+vendor: RADAR
+model: aRMT-App
+version: 1.17.0
+assessment_type: QUESTIONNAIRE
+doc: aRMT Questionnaires definition. Includes Personal Health Questionnaire Depression Scale (PHQ-8), Experience sampling method (ESM) and RSES and other data.
+data:
+  - type: THINC_IT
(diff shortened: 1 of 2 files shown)


mpgxvii (approved):
This is a good idea. LGTM. Thanks!

specifications/active/aRMT-1.17.0.yml
```diff
@@ -0,0 +1,299 @@
+name: aRMT
```
nivemaham, line 1: why do we need this new specification?

blootsvoets: It removes the topics that are already in the general spec. There isn’t a conflict though, so I can also remove this and just keep 1.16.0…

nivemaham: Should it be `questionnaire_response` instead of `notification_thinc_it` and so on in that case?

blootsvoets: Well, so the generic topics, including `questionnaire_response` are now in the generic specification, so I thought there would be no need to duplicate them here.

PR #324, opened 2023-01-13
new property macAddressHash has been added

this-Aditya:


this-Aditya:
This pr is in accordance to https://github.com/RADAR-base/radar-commons-android/pull/389

blootsvoets:
Hi, to send the MAC address, a new schema file needs to be created in the same directory, which will be sent to a new topic. Only the time, timeReceived and macAddressHash fields are needed in that schema. You can copy time and timeReceived fields from this schema. Record name would be PhoneBluetoothDeviceScanned.

this-Aditya:
sure! do i need to remove macAddressHash field from this schema

commons/passive/phone/phone_bluetooth_devices_scanned.avsc
```diff
@@ -0,0 +1,11 @@
+{
+  "namespace": "org.radarcns.passive.phone",
+  "type": "record",
+  "name": "PhoneBluetoothDevicesScanned",
+  "doc": "Phone Bluetooth devices info.",
+  "fields": [
+    {"name": "time", "type": "double", "doc": "Device timestamp in UTC (s)."},
+    {"name": "timeReceived", "type": "double", "doc": "Device receiver timestamp in UTC (s)."},
+    {"name": "macAddressHash", "type": ["null", "bytearray"], "default": null, "doc":"Hash of Nearby Bluetooth device MAC address"}
```
afolarin, line 9: If we are taking a snapshot of the Bluetooth environment then presumable we should have a set of hashes (so probably pluralise the doc and name?
```suggestion
    {"name": "macAddressHashes", "type": ["null", "bytearray"], "default": null, "doc":"Hash of Nearby Bluetooth device MAC addresses"}
```

afolarin: thinking again perhaps it is better not to collect the set of devices in a single payload for the snapshot?

yatharthranjan: I believe the plan is to send this record for each bluetooth device. That's why it was separated from the bluetooth schema.

afolarin: Yep, I saw Joris's comment after I put that in. I agree it's a better proposition.

this-Aditya: So, do I make these changes or not?

yatharthranjan: No, should not be needed.

afolarin:
@blootsvoets @yatharthranjan any thoughts on merging this schema?

blootsvoets:
One more thought, along with the hash salt itself, the android program should store a random identifier. That random identifier should be sent in this record. That way, if the app gets reinstalled or installed on another device, it's clear during analysis that the mac addresses between iterations are not comparable. See for example
```
https://github.com/RADAR-base/RADAR-Schemas/blob/45d8c8ae4ddb997966b8ef9d98d98b8bec4a104b/commons/passive/phone/phone_relative_location.avsc#L9
```
In this case, it could be called `hashSaltReference`.

this-Aditya:
As once I was facing some issues and deleted its forked repository. Now I lost the reference to this pull request and I am not able to edit the changes. That's why I am creating a new pull request which includes the changes as needed.

this-Aditya:
Here it is https://github.com/RADAR-base/RADAR-Schemas/pull/330.

PR #336, opened 2023-05-27
Schemas added for google sleep APIs

this-Aditya:
Added schemas and specifications required for google sleep api.

### DIFF
--- a/commons/passive/google/google_sleep_classify_event.avsc
+++ b/commons/passive/google/google_sleep_classify_event.avsc
@@ -2,12 +2,12 @@
   "namespace": "org.radarcns.passive.google",
   "type": "record",
   "name": "GoogleSleepClassifyEvent",
-  "doc": "Sleep classification event including the sleep confidence, device motion and ambient light level.",
+  "doc": "Sleep classification event that indicates that the user is probably sleeping. Includes sleep confidence, device motion and ambient light level.",
   "fields": [
     { "name": "time", "type": "double", "doc": "Device timestamp in UTC (s)." },
     { "name": "timeReceived", "type": "double", "doc": "Device receiver timestamp in UTC (s)." },
-    { "name": "sleepConfidence", "type": "int", "doc": "Sleep confidence value between 0 and 100." },
-    { "name": "light", "type": "int", "doc": "The brightness of the space around the device. Value ranges from 1 to 6, inclusive." },
-    { "name": "motion", "type": "int", "doc": "The amount of device motion. Value ranges from 1 to 6, inclusive." }
+    { "name": "sleepConfidence", "type": "int", "doc": "Sleep confidence value. Higher values indicate that the user is more likely sleeping, while lower values indicate that the user is more likely awake." },
+    { "name": "light", "type": "int", "doc": "The brightness of the space around the device, based on the device's ambient light sensor readings. Value ranges from 1 to 6, inclusive. Higher values indicate brighter surroundings, while lower values indicate darker surroundings." },
+    { "name": "motion", "type": "int", "doc": "The amount of device motion, based on the device's accelerometer readings. Value ranges from 1 to 6, inclusive. Higher values indicate more movement of the device." }
   ]
 }
--- a/commons/passive/google/google_sleep_classify_event.avsc
+++ b/commons/passive/google/google_sleep_classify_event.avsc
@@ -1,0 +1,13 @@
+{
+  "namespace": "org.radarcns.passive.google",
+  "type": "record",
+  "name": "GoogleSleepClassifyEvent",
+  "doc": "Sleep classification event including the sleep confidence, device motion and ambient light level.",
+  "fields": [
+    { "name": "time", "type": "double", "doc": "Device timestamp in UTC (s)." },
+    { "name": "timeReceived", "type": "double", "doc": "Device receiver timestamp in UTC (s)." },
+    { "name": "sleepConfidence", "type": "int", "doc": "Sleep confidence value between 0 and 100." },
+    { "name": "light", "type": "int", "doc": "The brightness of the space around the device. Value ranges from 1 to 6, inclusive." },
+    { "name": "motion", "type": "int", "doc": "The amount of device motion. Value ranges from 1 to 6, inclusive." }
+  ]
+}
--- a/commons/passive/google/google_sleep_classify_event.avsc
+++ b/commons/passive/google/google_sleep_classify_event.avsc
@@ -6,7 +6,7 @@
   "fields": [
     { "name": "time", "type": "double", "doc": "Device timestamp in UTC (s)." },
     { "name": "timeReceived", "type": "double", "doc": "Device receiver timestamp in UTC (s)." },
-    { "name": "sleepConfidence", "type": "float", "doc": "Sleep confidence value. Higher value indicate that the user is more likely sleeping, while lower values indicate that the user is more likely awake." },
+    { "name": "sleepConfidence", "type": "float", "doc": "Sleep confidence value between 0-1. Higher value indicate that the user is more likely sleeping, while lower values indicate that the user is more likely awake." },
     { "name": "light", "type": "int", "doc": "The brightness of the space around the device, based on the device's ambient light sensor readings. Value ranges from 1 to 6, inclusive. Higher values indicate brighter surroundings, while lower values indicate darker surroundings." },
     { "name": "motion", "type": "int", "doc": "The amount of device motion, based on the device's accelerometer readings. Value ranges from 1 to 6, inclusive. Higher values indicate more movement of the device." }
   ]
--- a/commons/passive/google/google_sleep_classify_event.avsc
+++ b/commons/passive/google/google_sleep_classify_event.avsc
@@ -6,7 +6,7 @@
   "fields": [
     { "name": "time", "type": "double", "doc": "Device timestamp in UTC (s)." },
     { "name": "timeReceived", "type": "double", "doc": "Device receiver timestamp in UTC (s)." },
-    { "name": "sleepConfidence", "type": "int", "doc": "Sleep confidence value. Higher values indicate that the user is more likely sleeping, while lower values indicate that the user is more likely awake." },
+    { "name": "sleepConfidence", "type": "float", "doc": "Sleep confidence value. Higher value indicate that the user is more likely sleeping, while lower values indicate that the user is more likely awake." },
     { "name": "light", "type": "int", "doc": "The brightness of the space around the device, based on the device's ambient light sensor readings. Value ranges from 1 to 6, inclusive. Higher values indicate brighter surroundings, while lower values indicate darker surroundings." },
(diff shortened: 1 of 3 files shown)


commons/passive/google/google_sleep_classify_event.avsc
```diff
@@ -0,0 +1,13 @@
...
+  "namespace": "org.radarcns.passive.google",
+  "type": "record",
+  "name": "GoogleSleepClassifyEvent",
+  "doc": "Sleep classification event including the sleep confidence, device motion and ambient light level.",
+  "fields": [
+    { "name": "time", "type": "double", "doc": "Device timestamp in UTC (s)." },
+    { "name": "timeReceived", "type": "double", "doc": "Device receiver timestamp in UTC (s)." },
+    { "name": "sleepConfidence", "type": "int", "doc": "Sleep confidence value between 0 and 100." },
+    { "name": "light", "type": "int", "doc": "The brightness of the space around the device. Value ranges from 1 to 6, inclusive." },
+    { "name": "motion", "type": "int", "doc": "The amount of device motion. Value ranges from 1 to 6, inclusive." }
```
blootsvoets, line 11: Are there any interpretations for these values 1 through 6? If so, can you please add them? Or a link to the interpretation?

commons/passive/google/google_sleep_classify_event.avsc
```diff
@@ -0,0 +1,13 @@
+{
+  "namespace": "org.radarcns.passive.google",
+  "type": "record",
+  "name": "GoogleSleepClassifyEvent",
+  "doc": "Sleep classification event including the sleep confidence, device motion and ambient light level.",
+  "fields": [
+    { "name": "time", "type": "double", "doc": "Device timestamp in UTC (s)." },
+    { "name": "timeReceived", "type": "double", "doc": "Device receiver timestamp in UTC (s)." },
+    { "name": "sleepConfidence", "type": "int", "doc": "Sleep confidence value between 0 and 100." },
```
blootsvoets, line 9: you can remove the `sleep` prefix. I'm not sure what the confidence means. Confidence of what? In general, for probability values please use a float with range 0 - 1.

commons/passive/google/google_sleep_classify_event.avsc
```diff
@@ -0,0 +1,13 @@
+{
+  "namespace": "org.radarcns.passive.google",
+  "type": "record",
+  "name": "GoogleSleepClassifyEvent",
+  "doc": "Sleep classification event including the sleep confidence, device motion and ambient light level.",
```
blootsvoets, line 5: From this description, I'm not sure what is being classified, e.g. what a classify event record would tell me. Is it connected to the segment event? If so, it would be helpful to have a common identifier, e.g. `segmentId`.

commons/passive/google/google_sleep_segment_event.avsc
```diff
@@ -0,0 +1,19 @@
+{
+  "namespace": "org.radarcns.passive.google",
+  "type": "record",
+  "name": "GoogleSleepSegmentEvent",
+  "doc": "Represents the result of sleep data after the user is awake.",
+  "fields": [
+    { "name": "time", "type": "double", "doc": "Device timestamp in UTC (s)." },
+    { "name": "timeReceived", "type": "double", "doc": "Device receiver timestamp in UTC (s)." },
+    { "name": "sleepStartTime", "type": "double", "doc": "The UNIX epoch time (s) for the moment when the user goes to sleep." },
```
blootsvoets, line 9: You can use the `time` variable for `startTime` instead. Then you can remove this field. In the doc of time, note that the time is when the user goes to sleep.

commons/passive/google/google_sleep_segment_event.avsc
```diff
@@ -0,0 +1,19 @@
...
+    { "name": "sleepEndTime", "type": "double", "doc": "The UNIX epoch time (s) for the moment when the user wakes up." },
+    { "name": "sleepDuration", "type": "double", "doc": "The amount of elapsed time (s), that the user was asleep." },
```
blootsvoets, line 11: Is the duration the same as `endTime - time`? If so, you can remove the sleepDuration variable. If it is different, please put in the doc what the difference is and rename to `durationAsleep`.

commons/passive/google/google_sleep_segment_event.avsc
```diff
@@ -0,0 +1,19 @@
+{
+  "namespace": "org.radarcns.passive.google",
+  "type": "record",
+  "name": "GoogleSleepSegmentEvent",
+  "doc": "Represents the result of sleep data after the user is awake.",
+  "fields": [
+    { "name": "time", "type": "double", "doc": "Device timestamp in UTC (s)." },
+    { "name": "timeReceived", "type": "double", "doc": "Device receiver timestamp in UTC (s)." },
+    { "name": "sleepStartTime", "type": "double", "doc": "The UNIX epoch time (s) for the moment when the user goes to sleep." },
+    { "name": "sleepEndTime", "type": "double", "doc": "The UNIX epoch time (s) for the moment when the user wakes up." },
```
blootsvoets, line 10: Please rename to `endTime`

commons/passive/google/google_sleep_segment_event.avsc
```diff
@@ -0,0 +1,19 @@
...
+    { "name": "sleepDuration", "type": "double", "doc": "The amount of elapsed time (s), that the user was asleep." },
+    { "name": "sleepStatus", "type": {
```
blootsvoets, line 12: Please rename to `status`

commons/passive/google/google_sleep_segment_event.avsc
```diff
@@ -0,0 +1,19 @@
...
+      "name": "SleepStatus",
```
blootsvoets, line 13: Rename to `SleepClassificationStatus`

commons/passive/google/google_sleep_segment_event.avsc
```diff
@@ -0,0 +1,19 @@
...
+      "type": "enum",
+      "doc": "The status of the sleep segment detection.",
+      "symbols": ["STATUS_SUCCESSFUL", "STATUS_MISSING_DATA", "STATUS_NOT_DETECTED", "UNKNOWN"]
```
blootsvoets, line 16: Rename symbols to `SUCCESSFUL`, `MISSING_DATA`, `NOT_DETECTED` and `UNKNOWN`.

commons/passive/google/google_sleep_segment_event.avsc
```diff
@@ -0,0 +1,19 @@
...
+    }, "doc": "Status that indicates whether the system succeeded in detecting sleep.", "default": "UNKNOWN" }
```
blootsvoets, line 17: Please elaborate for each symbol what it means.

blootsvoets (commented):
Thanks for the schemas. They look complete, however I do have some questions and suggested some changes.

commons/passive/google/google_sleep_classify_event.avsc
this-Aditya: Yes, here it is https://developers.google.com/android/reference/com/google/android/gms/location/SleepClassifyEvent#:~:text=public%20int%20getLight,of%20the%20device.

commons/passive/google/google_sleep_classify_event.avsc
this-Aditya: Actually, here confidence means sleep confidence as mentioned [here](https://developers.google.com/android/reference/com/google/android/gms/location/SleepClassifyEvent#:~:text=public%20int-,getConfidence,-()).

commons/passive/google/google_sleep_classify_event.avsc
this-Aditya: Hi, classified events serve as supporting data for segmented events that are reported at regular intervals. In contrast, segmented events are only reported when there is sufficient confidence that the user is awake. As for the relationship between segmented and classified events, the last segmented event can also be reported whenever we subscribe to sleep data, while classified events are independently reported at fixed intervals. Despite this, is there still a need for the segmentId

this-Aditya: The sleep detection result depends on sampled device motion and ambient light readings during the past day. For the customized sleep segmentation we are using Sleep Classify events.

this-Aditya: Yes, the segmentId can be used as sleep segment data, depending on the sleep classification data from the previous day. It works in a way that segmentId will increase every time SleepSegmentEvents are recorded. Would you like this or something different?

commons/passive/google/google_sleep_classify_event.avsc
blootsvoets: Ah OK. Can you please add the full docstring in the link to here (except the range)? Because this provides a reason of why it is called sleepConfidence.

commons/passive/google/google_sleep_classify_event.avsc
blootsvoets: Please add the interpretation to the doc, copy from the reference.

commons/passive/google/google_sleep_classify_event.avsc
blootsvoets: You don't need to use segmentId in that case, I didn't understand from the description what confidence meant. In the main doc, please do add something like `Sleep classification event that indicates that the user is probably sleeping. Includes sleep confidence, device motion and ambient light level.`

commons/passive/google/google_sleep_segment_event.avsc
blootsvoets: OK if it is the same, please remove.

commons/passive/google/google_sleep_classify_event.avsc
this-Aditya: Okay

this-Aditya:
Hi, I have made the suggested changes and will be making the changes in the radar-commons-android as well.

commons/passive/google/google_sleep_classify_event.avsc
```diff
@@ -0,0 +1,13 @@
+{
+  "namespace": "org.radarcns.passive.google",
+  "type": "record",
+  "name": "GoogleSleepClassifyEvent",
+  "doc": "Sleep classification event that indicates that the user is probably sleeping. Includes sleep confidence, device motion and ambient light level.",
+  "fields": [
+    { "name": "time", "type": "double", "doc": "Device timestamp in UTC (s)." },
+    { "name": "timeReceived", "type": "double", "doc": "Device receiver timestamp in UTC (s)." },
+    { "name": "sleepConfidence", "type": "float", "doc": "Sleep confidence value. Higher value indicate that the user is more likely sleeping, while lower values indicate that the user is more likely awake." },
```
blootsvoets, line 9: Could you just add the range 0 - 1 to this doc?

blootsvoets (approved):
After these changes it's really good and understandable. Thank you for your efforts!

this-Aditya:
You're welcome! Thank you for your appreciation!

PR #339, opened 2023-06-08
Fix Healthkit schema

mpgxvii:
- Update `time` field in Healthkit generic schema (change `startTime` to `time`) to fix issues with the filename in output restructure


yatharthranjan (approved):
LGTM

PR #346, opened 2023-07-20
removed unsupported activity types from activity transition event

this-Aditya:
During the testing of the activity transition plugin, events for `TILTING` and `UNKNOWN` are not supported. The plugin throws an exception with the message `ActivityTransitionRequest specified an unsupported transition activity type`.

### DIFF
--- a/commons/passive/google/google_activity_transition_event.avsc
+++ b/commons/passive/google/google_activity_transition_event.avsc
@@ -9,8 +9,8 @@
     { "name": "activity", "type": {
       "name": "ActivityType",
       "type": "enum",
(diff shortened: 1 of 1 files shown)


mpgxvii (commented):
@this-Aditya Are the `TILTING` and `UNKNOWN` events in the docs?

this-Aditya:
For that error, I found [this](https://stackoverflow.com/questions/49475348/invalid-activities-specified-in-the-activitytransitionrequest) on Stack Overflow.

this-Aditya:
@mpgxvii Till now, I have not received any events from, `ON-FOOT` but it is not throwing any exception while registering.

PR #347, opened 2023-07-26
Updated default value of activity to null.

this-Aditya:
Updated default value for an activity to null in `google_activity_transition_event.avsc`.

### DIFF
--- a/commons/passive/google/google_activity_transition_event.avsc
+++ b/commons/passive/google/google_activity_transition_event.avsc
@@ -6,12 +6,12 @@
   "fields": [
     { "name": "time", "type": "double", "doc": "The UNIX epoch time (s) for the moment when the activity takes place.." },
     { "name": "timeReceived", "type": "double", "doc": "Device receiver timestamp in UTC (s)." },
-    { "name": "activity", "type": {
+    { "name": "activity", "type": ["null",{
       "name": "ActivityType",
       "type": "enum",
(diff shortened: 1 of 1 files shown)


yatharthranjan (approved):
LGTM, is the same change required in sleep schema too?

this-Aditya:
@yatharthranjan Regarding the sleep data, the sleep API is not pushing any data of type `UNKNOWN` like what is happening here. It only provides data for the types defined in the documentation, which are `STATUS_SUCCESSFUL`, `STATUS_NOT_DETECTED`, and `STATUS_MISSING_DATA`, this might not be needed there.

this-Aditya:
HI, even if the Sleep API is not pushing `UNKNOWN`.  It is used in the else statement [here](https://github.com/RADAR-base/radar-commons-android/blob/077d77b092b1d655bff0e60cd4fa0594e9f3342e/plugins/radar-android-google-sleep/src/main/java/org.radarbase.passive.google.sleep/GoogleSleepManager.kt#L170C1-L171C1). The value passed from sleep API falls in these three only. Should this change be needed in the sleep schema too?

this-Aditya:
Hi Yatharth, Sorry for the oversight. In the case of the sleep plugin, the same scenario is taking place, thus changes are also needed there. I have updated that as well [here](https://github.com/RADAR-base/RADAR-Schemas/pull/348).

PR #348, opened 2023-07-29
Updated default value of sleep segment status to null. 

this-Aditya:
Updated schema `google_sleep_segement_event` to have default value of sleep status as null.


yatharthranjan (approved):
Lgtm

PR #362, opened 2024-01-25
Using PairState enum for paired status of remote bluetooth device.

this-Aditya:
Earlier the pairing state of the remote device is calculated manually, but there is also a straightforward and precise way [here](https://developer.android.com/reference/android/bluetooth/BluetoothDevice#getBondState()).


yatharthranjan (approved):
Thanks, LGTM.

Bdegraaf1234:
Merging to include in 0.8.7 release

PR #374, opened 2024-05-07
Release 0.8.8

mpgxvii:
- Bump to minor release instead of hotfix release


mpgxvii:
@Bdegraaf1234 I think we need to bump the version here? Since it is not able to release the artifact to the sonatype repository.

Bdegraaf1234 (approved):
Great idea, yes :)

PR #375, opened 2024-05-23
Replacing the use of deprecated Places.Type

this-Aditya:
The Google Places API Places.Type is [deprecated](https://developers.google.com/maps/documentation/places/android-sdk/reference/com/google/android/libraries/places/api/model/Place.Type#:~:text=This%20enum%20is%20deprecated.%0AUse%20Place.getPlaceTypes()%20instead.) in recent versions. Instead of using an enum for Places.Type, it now sends a list of strings for places.


yatharthranjan (approved):
LGTM

Issue #368, opened 2024-03-18
Confluence cloud interop broken


Bdegraaf1234:
When upgrading from release v0.8.5 to v0.8.6, the catalog-server:kafka-init container is unable to connect to confluent cloud.

The error occurs during authentication:

``` logs
[2024-03-18 11:28:22,586] ERROR - Schema registry ${confluence_cloud_url} not ready, responded with HTTP 401 Unauthorized: REST call to <${confluence_cloud_url}/subjects> failed (HTTP code 401 Unauthorized): {"error_code":401,"message":"Unauthorized"} (Emitters.kt:231)
```

The problem seems to originate [here](https://github.com/RADAR-base/RADAR-Schemas/blob/master/java-sdk/radar-schemas-registration/src/main/java/org/radarbase/schema/registration/SchemaRegistry.kt#L63). I believe the issue was introduced in this commit aefea95bd3c0be1c92cf282bdd31ffdcecdf4229

Debugging so far:
- Verified that the credentials are available to the pod
- Verified that manually providing these credentials to the url specified (${confluence_cloud_url}/subjects) works

Bdegraaf1234:
Solved by #369

PR #376, opened 2024-05-23
Fix schemaregistry

Bdegraaf1234:
Issues with basic authentication (see #368)

It seems that httpclient configuration was not properly passed to the `SchemaRegistry`

This PR is dependent on a new release of radar-commons


Bdegraaf1234:
dependent on: https://github.com/RADAR-base/radar-commons/pull/164

PR #382, opened 2024-07-17
Add Polar schemas

fschulting:
This PR adds Polar schemas for interaction with Polar devices. 
Tested with: Polar H10, Polar Vantage V3 and Polar Verity Sense


Bdegraaf1234 (approved):
LGTM

PR #352, opened 2023-08-01
Added schemas and specifications for data retrieved from Google Places API.

this-Aditya:
Changes:

Added `google_places_info` schema.
Created specification `google_places-1.0.0`, included specifications for google places API.

Fixes https://github.com/RADAR-base/RADAR-Schemas/issues/350

PR #386, opened 2024-09-09
Release 0.8.11

yatharthranjan:
- Add polar SDK schemas and specs
- update secret for snapshot publishing
- pRMT audio input schemas and spec
- Various security fixes


afolarin:
For the pRMT audio plugin, should we add button event markers to the schema? Or is this using something else?

yatharthranjan:
Hi @afolarin, I thought it could be separate from the audio plugin so it can be utilised with any plugin. It could be either its own plugin or part of the core application. 
We can also have a per-source event buttons (a button in each source in the table for marking events). The functionality of each can be delegated to the respective source plugin. By default, it can just send the time and event name for each.

afolarin:
Yes, it would be good to have a standard "_user-triggered event_" for RADAR-base generally. The semantics I guess could mean it is used anywhere (p/aRMT apps, wearables etc). I guess then it needs to have some capability to include some context, which would be useful as we could then specify as part of e.g. config what the event means in that specific use case/project etc.

yatharthranjan:
Ok sure, I have added an issue to track this -- https://github.com/RADAR-base/radar-prmt-android/issues/149

commons/passive/audio/phone_audio_input.avsc
```diff
@@ -0,0 +1,24 @@
...
+        { "name": "deviceSampleRates", "type": "string", "doc": "Supported sample rates of the input audio device." },
+        { "name": "deviceEncodings", "type": "string", "doc": "Supported encodings of the input audio device." },
+        { "name": "deviceType", "type": "string", "doc": "Type of the input audio device used for recording." },
+        { "name": "deviceChannelCounts", "type": "string", "doc": "Supported channel counts of the input audio device." },
+        { "name": "audioLength", "type": "long", "doc": "Length of the audio recording (in milliseconds)." },
+        { "name": "audioFileSize", "type": "long", "doc": "Size of the audio file (in bytes)." },
+        { "name": "hadPlayback", "type": "boolean", "doc": "Whether the recorded audio file was played before uploading to s3 storage." },
+        { "name": "audioFileExtension", "type": "string", "doc": "Extension of the audio file." },
+        { "name": "configuredSampleRate", "type": "int", "doc": "Sample rate for audio recording configured by firebase remote configs in application."},
```
mpgxvii, line 21: Will this always be int and no decimals?

yatharthranjan: I think so, but @this-Aditya can you confirm?

commons/passive/audio/phone_audio_input.avsc
```diff
@@ -0,0 +1,24 @@
...
+    "name": "PhoneAudioInput",
+    "doc": "Uncompressed high-quality audio data collected by the PhoneAudioInput plugin, making use of low-level classes that interact directly with hardware.",
+    "fields": [
+        { "name": "time", "type": "double", "doc": "Device timestamp in UTC (s)." },
+        { "name": "timeReceived", "type": "double", "doc": "Device receiver timestamp in UTC (s)." },
+        { "name": "fileName", "type": "string", "doc": "Name of the audio file after it is saved to S3." },
+        { "name": "filePath", "type": "string", "doc": "Path of the audio file retrieved after uploading to S3 storage." },
+        { "name": "deviceName", "type": "string", "doc": "Name of the input audio device used for routing during this recording." },
+        { "name": "deviceId", "type": "string", "doc": "Identifier associated with the input device used for audio recording." },
+        { "name": "deviceSampleRates", "type": "string", "doc": "Supported sample rates of the input audio device." },
```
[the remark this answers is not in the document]
yatharthranjan: good point, @this-Aditya is this supposed to be an array or will it be like a comma-separated string?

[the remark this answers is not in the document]
this-Aditya: Initially, I planned to use an array for this, but I remember [joris's comment](https://github.com/RADAR-base/RADAR-Schemas/pull/352#discussion_r1299812336) to avoid using an array here.

commons/passive/audio/phone_audio_input.avsc
this-Aditya: Yes, [it is](https://github.com/RADAR-base/radar-commons-android/blob/958721598f52d278f2eead88e094a26ce95c7185/plugins/radar-android-phone-audio-input/src/main/java/org/radarbase/passive/phone/audio/input/PhoneAudioInputManager.kt#L72).

mpgxvii: Ah ok I see, thanks!

pvannierop:
I am merging this release because every reviewers comment appears to have been addressed.

PR #387, opened 2024-10-04
Add support for Sentry monitoring

pvannierop:
Description: This PR will add option to monitor the application with Sentry. Sentry monitoring will be enabled by the following:

1. Set a `SENTRY_DSN` environment variable that points to the desired Sentry DSN.
2. (Optional) Set the `SENTRY_LOG_LEVEL` environment variable to control the minimum log level of events sent to Sentry.
   The default log level for Sentry is `WARN`. Possible values are `TRACE`, `DEBUG`, `INFO`, `WARN`, and `ERROR`.

The docs were updated with these instructions.

### DIFF
--- a/java-sdk/radar-catalog-server/build.gradle.kts
+++ b/java-sdk/radar-catalog-server/build.gradle.kts
@@ -1,3 +1,9 @@
+plugins {
+    // TODO Remove this when new release of radar-commons is available and used in this project.
+    // This version has Sentry support built in for radar-kotlin plugin.
+    id("io.sentry.jvm.gradle") version "4.11.0"
+}
+
 description = "RADAR Schemas specification and validation tools."
 
 dependencies {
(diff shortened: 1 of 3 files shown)


java-sdk/radar-catalog-server/build.gradle.kts
```diff
@@ -14,3 +21,7 @@ dependencies {
 application {
     mainClass.set("org.radarbase.schema.service.SourceCatalogueServer")
 }
+
+radarKotlin {
+    sentryEnabled.set(true)
```
ewelinagr, line 26: javaVersion.set(Versions.java)
    log4j2Version.set(Versions.log4j2)
    
    not needed?

pvannierop: @ewelinagr No, defaults are defined in the radar-commons kotlin project gradle plugin. Better to use these defaults.

PR #399, opened 2025-06-18
Removed hashSaltReference field from PhoneBluetoothDeviceScanned Schema

this-Aditya:


this-Aditya:
Hi, @yatharthranjan @mpgxvii, I'm not sure if this field needs to be deleted, or if we can simply send a null value for it without removing the field.

PR #426, opened 2026-06-26
Use null defaults for respiratory rate sleep summary data fields

this-Aditya:
I noticed that the data from `respiratory-rate-sleep-summary` is non-nullable for breaths across different sleep stages, those fields are marked as optional in the documentation. Currently, the converter defaults missing values to 0 instead of sending null or rejecting the data. This PR corrects that.


java-sdk/buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -1,5 +1,5 @@
 object Versions {
-    const val project = "0.8.17"
+    const val project = "0.8.18"
```
mpgxvii: Hi @this-Aditya, I think this file is not being used anymore. Would you be able to resolve the conflicts? Thank you.

this-Aditya: Resolved now.

mpgxvii (approved):
LGTM!

PR #431, opened 2026-07-30
Dexcom schema

Chin-eng:


this-Aditya:
Hi, @Chin-eng, I see that CI is failing because some schemas are missing the "doc" property. Please add the `doc` property to all fields where it’s absent so that CI passes?

Chin-eng:
> Hi, @Chin-eng, I see that CI is failing because some schemas are missing the "doc" property. Please add the `doc` property to all fields where it’s absent so that CI passes?

Done

this-Aditya:
Hi @Chin-eng, could you make the records flat where needed, as [the readme of this project](https://github.com/RADAR-base/RADAR-Schemas#:~:text=Prefer%20a%20flat%20record%20over%20a%20hierarchical%20record.%20This%20simplifies%20the%20organization%20of%20the%20data%20downstream%2C%20for%20example%2C%20when%20mapping%20to%20CSV.) also suggests to do that.

commons/connector/dexcom/dexcom_alert.avsc
```diff
@@ -0,0 +1,69 @@
...
+  "fields": [
+    {
+      "name": "recordId",
+      "type": "string",
+      "doc": "Dexcom record UUID."
+    },
+    {
+      "name": "systemTime",
+      "type": "double",
+      "doc": "System time of the alert in UTC (s). Parsed from Dexcom ISO 8601 systemTime."
```
this-Aditya: Could we make the `systemTime` to `time` here. Almost all our schemas use time as the main timestamp.

commons/connector/dexcom/dexcom_alert.avsc
```diff
@@ -0,0 +1,69 @@
...
+      "name": "displayTime",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Display time as ISO 8601. Receiver records may have no offset."
+    },
+    {
+      "name": "alertName",
+      "type": "string",
+      "doc": "Alert name (e.g. high, low, rise, fall, urgentLow)."
+    },
```
this-Aditya, line 28: `alertName` is a closed list in the API, could you make this an enum instead of a string, like: `["null", DexcomAlertName]` with default null, and keep `UNKNOWN`  as a fallback.

commons/connector/dexcom/dexcom_alert.avsc
```diff
@@ -0,0 +1,69 @@
...
+    {
+      "name": "alertState",
+      "type": "string",
+      "doc": "Alert state (e.g. activeAlarming, activeSnoozed, inactive)."
```
this-Aditya: Same as `alertName`. Could you make it `["null", DexcomAlertState]` with default null.

commons/connector/dexcom/dexcom_alert.avsc
```diff
@@ -0,0 +1,69 @@
...
+    },
+    {
+      "name": "displayDevice",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Display device type: iOS, Android, or receiver. Null if unknown."
+    },
```
this-Aditya: Same pattern `displayDevice` can be an enum too, the API only returns unknown, receiver, iOS, android or unknown.

commons/connector/dexcom/dexcom_alert.avsc
```diff
@@ -0,0 +1,69 @@
...
+      "name": "transmitterId",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Hashed Dexcom transmitter identifier. Null if unknown."
+    },
+    {
+      "name": "displayApp",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Display application name (for example G6, G7, or Follow). Null if unknown."
```
this-Aditya: The doc says `Follow` here but that is not one of the values Dexcom lists.

commons/connector/dexcom/dexcom_alert.avsc
```diff
@@ -0,0 +1,69 @@
+{
+  "namespace": "org.radarcns.connector.dexcom",
+  "type": "record",
+  "name": "DexcomAlert",
```
this-Aditya, line 4: Could you expand the docs in this file using the official reference at: https://developer.dexcom.com/docs/swaggerv3/other/getalertseventsv3.

commons/connector/dexcom/dexcom_alert.avsc
```diff
@@ -0,0 +1,69 @@
...
+    {
+      "name": "displayTime",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Display time as ISO 8601. Receiver records may have no offset."
+    },
+    {
+      "name": "alertName",
+      "type": "string",
+      "doc": "Alert name (e.g. high, low, rise, fall, urgentLow)."
```
this-Aditya: One thing worth adding is that Dexcom one+ does not support Urgent Low, Urgent Low Soon, Rising Fast or Falling Fast alerts.. Ref: [official docs](https://developer.dexcom.com/docs/swaggerv3/other/getalertseventsv3#:~:text=Dexcom%20ONE%2B%20does%20not%20support%20the%20following%20alerts%3A%20Urgent%20Low%2C%20Urgent%20Low%20Soon%2C%20Rising%20Fast%2C%20Falling%20Fast.)

commons/connector/dexcom/dexcom_egv.avsc
```diff
@@ -0,0 +1,100 @@
...
+    {
+      "name": "recordId",
+      "type": "string",
+      "doc": "Dexcom record UUID."
+    },
+    {
+      "name": "systemTime",
+      "type": "double",
+      "doc": "System time of the EGV in UTC (s). Parsed from Dexcom ISO 8601 systemTime."
+    },
```
this-Aditya, line 16: Same as on the alert schema, could we use time as how are we doing in all schemas.

commons/connector/dexcom/dexcom_egv.avsc
```diff
@@ -0,0 +1,100 @@
...
+      "name": "transmitterTicks",
+      "type": ["null", "long"],
+      "default": null,
+      "doc": "Seconds since transmitter initiation. Null if unknown."
+    },
+    {
+      "name": "value",
+      "type": ["null", "int"],
+      "default": null,
+      "doc": "EGV in mg/dL. 39 and 401 indicate out-of-display-range values."
```
this-Aditya: Could you take the full wording from the official reference here.
Roughly -> "....39 means the value was calculated as less than 40 and is used in calculations but not shown to end users, 40 to 400 is the actual measuring range of the CGM, and 401 means it was calculated as greater than 400 and again is not shown......."

commons/connector/dexcom/dexcom_egv.avsc
```diff
@@ -0,0 +1,100 @@
...
+    },
+    {
+      "name": "status",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "EGV status when value is special (e.g. low, high)."
```
this-Aditya: status is a closed list only "high" and "low". Could you make it an enum.

commons/connector/dexcom/dexcom_egv.avsc
```diff
@@ -0,0 +1,100 @@
...
+    },
+    {
+      "name": "trend",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Trend arrow (e.g. flat, singleUp)."
```
this-Aditya: Could you make trend an enum as well.

commons/connector/dexcom/dexcom_egv.avsc
```diff
@@ -0,0 +1,100 @@
...
+      "name": "rateUnit",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Unit for trendRate (e.g. mg/dL/min)."
+    },
+    {
+      "name": "displayDevice",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Display device type: iOS, Android, or receiver. Null if unknown."
```
this-Aditya: `displayDevice` can use the same enum I've talked about on `dexcom_alert.avsc`.

commons/connector/dexcom/dexcom_egv.avsc
```diff
@@ -0,0 +1,100 @@
...
+      "name": "unit",
+      "type": "string",
+      "doc": "Unit of value (mg/dL or mmol/L)."
+    },
+    {
+      "name": "rateUnit",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Unit for trendRate (e.g. mg/dL/min)."
+    },
```
this-Aditya, line 89: unit is a required string here but rateUnit is nullable, and in the API both are required. See [here](https://developer.dexcom.com/docs/swaggerv3/other/getestimatedglucosevaluesv3#:~:text=up%20or%20down.-,records%5B%5D.%E2%80%8Bunit,such%20as%20mmol/L%2C%20is%20the%20responsibility%20of%20the%20client%20application.,-records%5B%5D.%E2%80%8BdisplayDevice). Also worth noting in the doc that Dexcom always send mg/dL

commons/connector/dexcom/dexcom_calibration.avsc
```diff
@@ -0,0 +1,71 @@
+{
+  "namespace": "org.radarcns.connector.dexcom",
+  "type": "record",
+  "name": "DexcomCalibration",
+  "doc": "Calibration event from Dexcom CGM API v3.",
```
this-Aditya: It it worth mentioning that dexcom one does not support the calibrations endpoint t all and returns an empty array. 
[See the docs](https://developer.dexcom.com/docs/swaggerv3/other/getcalibrationsv3#:~:text=Dexcom%20ONE%20does%20not%20support%20the%20/calibrations%20endpoint%20and%20will%20return%20an%20empty%20array.)

commons/connector/dexcom/dexcom_calibration.avsc
```diff
@@ -0,0 +1,71 @@
...
+      "name": "unit",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Unit of value (mg/dL or mmol/L)."
+    },
+    {
+      "name": "value",
+      "type": ["null", "int"],
+      "default": null,
+      "doc": "Fingerstick calibration glucose value."
```
this-Aditya: Could you correct this doc like: "this is a blood glucose value the user measured on a normal meter and typed in by hand..". Or from the [api docs](https://developer.dexcom.com/docs/swaggerv3/other/getcalibrationsv3).

commons/connector/dexcom/dexcom_calibration.avsc
```diff
@@ -0,0 +1,71 @@
...
+    },
+    {
+      "name": "displayDevice",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Display device type: iOS, Android, or receiver. Null if unknown."
```
this-Aditya: `displayDevice` can use the same enum I've talked about on dexcom_alert.avsc.

commons/connector/dexcom/dexcom_calibration.avsc
```diff
@@ -0,0 +1,71 @@
...
+  "fields": [
+    {
+      "name": "recordId",
+      "type": "string",
+      "doc": "Dexcom record UUID."
+    },
+    {
+      "name": "systemTime",
+      "type": "double",
+      "doc": "System time of the calibration in UTC (s). Parsed from Dexcom ISO 8601 systemTime."
```
this-Aditya: Similarly this can also be time.

commons/connector/dexcom/dexcom_event.avsc
```diff
@@ -0,0 +1,81 @@
+{
+  "namespace": "org.radarcns.connector.dexcom",
+  "type": "record",
+  "name": "DexcomEvent",
+  "doc": "User-entered event from Dexcom CGM API v3 (insulin, carbs, exercise, health, etc.).",
+  "fields": [
```
this-Aditya, line 6: The API also returns [recordedSystemTime](https://developer.dexcom.com/docs/swaggerv3/other/geteventsv3#:~:text=records%5B%5D.%E2%80%8BrecordedSystemTime) and [recordedDisplayTime](https://developer.dexcom.com/docs/swaggerv3/other/geteventsv3#:~:text=records%5B%5D.%E2%80%8BrecordedDisplayTime) and I do not see them here. Could you add both. They are not similar to systemTime and the displayTime, but they are when the user actually logged the entry as opposed to when it happened, so someone entering their morning info in the evening shows up as a gap between the two.

commons/connector/dexcom/dexcom_event.avsc
```diff
@@ -0,0 +1,81 @@
...
+    },
+    {
+      "name": "eventStatus",
+      "type": "string",
+      "doc": "Record status (created, updated, deleted)."
+    },
+    {
+      "name": "eventType",
+      "type": "string",
+      "doc": "Event type (insulin, carbs, exercise, health, bloodGlucose, notes)."
```
this-Aditya: This could be enum. As the [doc also list that](https://developer.dexcom.com/docs/swaggerv3/other/geteventsv3#:~:text=of%20the%20event.-,Enum,%22notes%22,-records%5B%5D.%E2%80%8BeventSubType)

commons/connector/dexcom/dexcom_event.avsc
```diff
@@ -0,0 +1,81 @@
...
+      "name": "displayTime",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Display time as ISO 8601 with offset when from mobile app."
+    },
+    {
+      "name": "eventStatus",
+      "type": "string",
+      "doc": "Record status (created, updated, deleted)."
+    },
```
this-Aditya, line 36: This can also be enum, [see docs](https://developer.dexcom.com/docs/swaggerv3/other/geteventsv3#:~:text=by%20the%20user.-,Enum,%22deleted%22,-records%5B%5D.%E2%80%8BeventType).

commons/connector/dexcom/dexcom_event.avsc
```diff
@@ -0,0 +1,81 @@
...
+      "name": "eventType",
+      "type": "string",
+      "doc": "Event type (insulin, carbs, exercise, health, bloodGlucose, notes)."
+    },
+    {
+      "name": "eventSubType",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Subtype (e.g. longActing, fastActing for insulin)."
+    },
```
this-Aditya: The API also send eventSubType as enum, [see docs](https://developer.dexcom.com/docs/swaggerv3/other/geteventsv3#:~:text=eventType%20%22bloodGlucose%22%3A%20null-,Enum,%22highSymptoms%22,-%2B3%20more).

commons/connector/dexcom/dexcom_event.avsc
this-Aditya: Could you expand this field doc a bit, something like: "Users can edit and delete their own entries and Dexcom keep sending us the revisions, so a record with status updated or deleted supersedes an earlier one with the same recordId...." 
Also [see doc](https://developer.dexcom.com/docs/swaggerv3/other/geteventsv3#:~:text=Describes%20the%20status%20of%20the%20record%20being%20posted%20from%20the%20health%20database.%20It%20describes%20if%20it%20is%20a%20brand%2Dnew%20entry%2C%20or%20if%20the%20record%20has%20been%20modified%2C%20or%20if%20the%20record%20has%20been%20deleted%20by%20the%20user.)

commons/connector/dexcom/dexcom_event.avsc
```diff
@@ -0,0 +1,81 @@
...
+      "name": "transmitterGenerationVariant",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Transmitter generation variant. Null if unknown."
+    },
+    {
+      "name": "displayDevice",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Display device type: iOS, Android, or receiver. Null if unknown."
```
this-Aditya: Again, this can be enum, and can be reusable, see my other comments on this field on other schemas.

commons/connector/dexcom/dexcom_alert.avsc
```diff
@@ -0,0 +1,69 @@
...
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Display device type: iOS, Android, or receiver. Null if unknown."
+    },
+    {
+      "name": "transmitterGeneration",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Transmitter generation (for example g6 or g7). Null if unknown."
+    },
```
this-Aditya: The API mentions this as [required](https://developer.dexcom.com/docs/swaggerv3/other/getalertseventsv3#:~:text=unknown%22%20%22receiver%22%20%22iOS%22%20%22android%22-,records%5B%5D.%E2%80%8BtransmitterGeneration,required,-The%20generation%20of%20transmitter) field, then this should also be non nullable for us.  Same check this in other schemas too.

commons/connector/dexcom/dexcom_alert.avsc
```diff
@@ -0,0 +1,69 @@
...
+    {
+      "name": "transmitterGenerationVariant",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Transmitter generation variant. Null if unknown."
+    },
```
this-Aditya: Please make this non null, as docs mention this as a required field.

commons/connector/dexcom/dexcom_alert.avsc
```diff
@@ -0,0 +1,69 @@
...
+    {
+      "name": "transmitterId",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Hashed Dexcom transmitter identifier. Null if unknown."
```
this-Aditya: Same, make it non null.

commons/connector/dexcom/dexcom_data_range.avsc
```diff
@@ -0,0 +1,196 @@
+{
+  "namespace": "org.radarcns.connector.dexcom",
+  "type": "record",
+  "name": "DexcomDataRange",
+  "doc": "Earliest and latest available times per data type from Dexcom API v3 /dataRange.",
```
this-Aditya: Before we go further with flattening this one, I am not sure this should be a topic at all. It is just saying what data the connector could fetch. I see it as a helper for syncing, so it is more connector state than participant data. That is also why this schema has no time field at the top level. @yatharthranjan can you confirm.

If we are keeping it, then please update the structure. start and end are the same two fields everywhere in this file, but they are written out six times under different names (`DexcomDataRangeMoment`, `DexcomDataRangeMomentEnd` .....). I think it would be better to define one record called `DexcomDataRangeWindow`, with start and end both typed as a single `DexcomDataRangeMoment` record holding systemTime and displayTime. So the hierarchy would be `DexcomDataRangeWindow` containing start and end, each one a `DexcomDataRangeMoment`.

commons/connector/dexcom/dexcom_device.avsc
```diff
@@ -0,0 +1,214 @@
...
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Transmitter generation variant. Null if unknown."
+    },
+    {
+      "name": "displayDevice",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Display device type: iOS, Android, or receiver. Null if unknown."
+    },
```
this-Aditya, line 22: Make this enum from the reusable record type I mentioned above.

commons/connector/dexcom/dexcom_device.avsc
```diff
@@ -0,0 +1,214 @@
+{
+  "namespace": "org.radarcns.connector.dexcom",
+  "type": "record",
+  "name": "DexcomDevice",
+  "doc": "Device information and alert settings from Dexcom CGM API v3.",
```
this-Aditya: I do not think this one can be made properly flat, a device has several schedules and each schedule has several alert settings, so the nesting is fine for this record.

commons/connector/dexcom/dexcom_device.avsc
```diff
@@ -0,0 +1,214 @@
...
+                      "default": null,
+                      "doc": "Snooze duration in minutes. Null if unknown."
+                    },
+                    {
+                      "name": "enabled",
+                      "type": "boolean",
+                      "doc": "Whether this alert is enabled."
+                    },
+                    {
+                      "name": "SecondaryTriggerCondition",
```
this-Aditya: this should be secondaryTriggerCondition with a lowercase s.

commons/connector/dexcom/dexcom_device.avsc
```diff
@@ -0,0 +1,214 @@
...
+                    "default": null,
+                    "doc": "Temporary override of this schedule. Null if none."
+                  },
+                  {
+                    "name": "daysOfWeek",
+                    "type": { "type": "array", "items": "string" },
+                    "doc": "Days of the week when this schedule applies."
+                  }
+                ]
+              },
```
this-Aditya: Could you flatten this one. Arrays map badly onto the sink connectors, this can be like: 

```json
There are only ever seven days, so this can just be seven boolean fields instead of an array:

{"name": "appliesOnSunday",    "type": ["null", "boolean"], "default": null...},
{"name": "appliesOnMonday",    ....},
{"name": "appliesOnTuesday",....
```

commons/connector/dexcom/dexcom_device.avsc
this-Aditya: For `dexcom_device` I think we can keep one topic but flatten it out. Right now it is four levels deep with two arrays, and arrays and nested records map badly onto the sink connectors. The [readme asks for flat records](https://github.com/RADAR-base/RADAR-Schemas/blob/49db3db7d360ff0702e1ed3ce379e1ad44181fa6/README.md?plain=1#L105) and [written out fields over arrays](https://github.com/RADAR-base/RADAR-Schemas/blob/49db3db7d360ff0702e1ed3ce379e1ad44181fa6/README.md?plain=1#L106) for the same reason.

What I would suggest is bringing the alert schedule fields up to the top level next to the device fields, doing the same with the three override fields, and turning daysOfWeek into seven booleans (I've added specific comment on that field, on how to do it). That leaves alertSettings as the only array, so move that up a level too and emit one record per alert setting rather than one per device. The device and schedule fields then just repeat on each row.


@yatharthranjan could you please confirm this hierarchy is fine.

this-Aditya (commented):
Hi @Chin-eng,  I've left comments inline could you please apply these

this-Aditya:
Hi @Chin-eng, could you please update [this PR](https://github.com/RADAR-base/RADAR-REST-Connector/pull/208) to reflect these schema updates once you’ve completed the updates for this PR?

this-Aditya:
Also, both of CI checks are failing, could you please check that?

Chin-eng:
okay can you run the workflow now

Chin-eng:
sorry I had accidentally put a g inside the syntax. can you run it one more time

Chin-eng:
I believe the CI crashed again because Dokka recursed until the stack overflowed. This might be happening because the same Avro types are declared in multiple files.

Instead of redeclaring the same enums, I referenced DexcomDisplayDevice, DexcomTransmitterGeneration, and DexcomTransmitterGenerationVariant from the other files and pushed the changes.

Could you run the workflow again and see if it passes now?

Chin-eng:
Okay since the stack overflow was not caused by declaring the same Avro types in each file, I have put those inline enums back. Schema validation should pass now.

commons/connector/dexcom/dexcom_event.avsc
```diff
@@ -0,0 +1,149 @@
...
+  "doc": "User-entered event from Dexcom CGM API v3 (insulin, carbs, exercise, health, etc.).",
+  "fields": [
+    {
+      "name": "recordId",
+      "type": "string",
+      "doc": "Dexcom record UUID."
+    },
+    {
+      "name": "systemTime",
+      "type": "double",
```
this-Aditya: Please rename this to time.

commons/connector/dexcom/dexcom_egv.avsc
```diff
@@ -0,0 +1,138 @@
+{
+  "namespace": "org.radarcns.connector.dexcom",
+  "type": "record",
+  "name": "DexcomEgv",
+  "doc": "Estimated glucose value (EGV) from Dexcom CGM API v3.",
```
this-Aditya, line 5: Please include the [trendRate field](https://developer.dexcom.com/docs/swaggerv3/other/getestimatedglucosevaluesv3#:~:text=records%5B%5D.%E2%80%8BtrendRate)

commons/connector/dexcom/dexcom_device.avsc
```diff
@@ -0,0 +1,210 @@
...
+    },
+    {
+      "name": "lastUploadDate",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Last upload timestamp as ISO 8601 UTC."
+    },
+    {
+      "name": "transmitterId",
+      "type": "string",
```
this-Aditya: Transmitter id can be null for egv and devices, but for other it is required ([see docs](https://developer.dexcom.com/docs/swaggerv3/other/getdevicesv3#:~:text=endDate%20query%20parameters.-,records%5B%5D.%E2%80%8BtransmitterId,string%20or%20null,-A%20hashed%20and)) better to make these null here.

commons/connector/dexcom/dexcom_egv.avsc
```diff
@@ -0,0 +1,138 @@
...
+    {
+      "name": "displayTime",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Display time as ISO 8601 with offset when from mobile app."
+    },
+    {
+      "name": "transmitterId",
+      "type": "string",
+      "doc": "A hashed and encrypted version of the unique identifier of the transmitter. This may be used to determine uniqueness of the transmitter or associate multiple records with the same transmitter, but it cannot be matched to the real transmitter identifier."
```
this-Aditya: Transmitter id can be null for egv and devices, but for other it is required ([see docs](https://developer.dexcom.com/docs/swaggerv3/other/getestimatedglucosevaluesv3#:~:text=have%20UTC%20offsets.-,records%5B%5D.%E2%80%8BtransmitterId,string%20or%20null,-A%20hashed%20and)) better to make these null here.

commons/connector/dexcom/dexcom_device.avsc
```diff
@@ -0,0 +1,210 @@
...
+    {
+      "name": "displayTime",
+      "type": ["null", "string"],
+      "default": null,
+      "doc": "Display time as ISO 8601. Null if unknown."
+    },
+    {
+      "name": "alertName",
+      "type": "string",
+      "doc": "Alert name this setting applies to."
```
this-Aditya: I think this can reuse the `DexcomAlertName` enum from dexcom_alert.avsc instead of a string.

commons/connector/dexcom/dexcom_alert.avsc
```diff
@@ -0,0 +1,88 @@
...
+  "fields": [
+    {
+      "name": "recordId",
+      "type": "string",
+      "doc": "A unique id (UUID) that identifies the alert."
+    },
+    {
+      "name": "time",
+      "type": "double",
+      "doc": "The recorded system time of the alerts/egv/calibration/event. This is the ISO 8601 time according to the system clock at which observation was made (UTC). Records sourced from mobile apps (with displayDevice of “iOS” or “android”) will have UTC offsets; records sourced from receivers (with displayDevice of “receiver”) will not have UTC offsets."
```
this-Aditya: Could you rewrite it to say it is the Dexcom systemTime converted to seconds since the Unix epoch, and same for other too.

this-Aditya (commented):
@Chin-eng, I've added some final comments, and after correcting these, could you please sync these to rest connector PR, so I can review that. 
Thanks

this-Aditya (approved):
LGTM!

Hi @mpgxvii @yatharthranjan, do you want to review this PR? Also, could you please confirm these memory configurations in the `build.gradle` and `gradle.properties` files here?

